PR #1, opened 2020-05-27
Replaced spa with hua

npr99:
The acronym spa was a hold over from an earlier version when the Housing Unit Allocation was called Stochastic Population Allocation.
spa needs to be replaced with hua, for consistency with the algorithm's name


npr99:
@ncsa-mo - I am testing the fork and pull request functions for pyincore. Let me know if this edit makes sense.

PR #2, opened 2020-06-29
Replace SPA acronym with HUA

npr99:
The SPA acronym was for Stochastic Population Allocation. The term was changed to Housing Unit Allocation. 
The correct acronym should be HUA


npr99:
@ncsa-mo - Hi Michael - I am trying to follow the steps for making a change to pyincore in Github.

ncsa-mo:
The change from SPA to HUA has been done in pyIncore 0.6.3 release.

PR #5, opened 2021-09-10
Pyincore util to convert population dislocation output for heatmap #3

longshuicy:
This utility method convert population dislocation output csv to a shapefile that only has `guid` and `numprec`. This will be used to post to data services to create a layer in geoserver, then further plot heatmap using the "numprec" field. 

When `filter_on` is true, it filter the csv to 1. has valid GUID 2. dislocation=TRue

To test: Run `/tests/pyincore/utils/test_popdisloutputprocessutil.py` to test


jonglee1:
this fixes #3

nunezco2 (approved):

PR #18, opened 2021-09-20
Added conda recipe

ywkim312:
Added conda build recipe

### DIFF
--- a/recipes/meta.yaml
+++ b/recipes/meta.yaml
@@ -1,0 +1,81 @@
+{% set name = "pyincore" %}
+{% set version = "1.0.0" %}
+ 
+package:
+  name: {{ name|lower }}
+  version: {{ version }}
+ 
+source:
+  path: /path/to/pyincore
+  # PR: https://opensource.ncsa.illinois.edu/bitbucket/projects/INCORE1/repos/pyincore/pull-requests/12/overview
+ 
+build:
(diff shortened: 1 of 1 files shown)


ywkim312:
Have a question about the python veraion. Is 3.6 okay or should be update it to higher version?

nunezco2 (approved):

recipes/meta.yaml
```diff
@@ -0,0 +1,81 @@
...
+ 
+test:
+  # Python imports
+  imports:
+    - pyincore
+    - pyomo
+   
+  requires:
+    # Put any additional test requirements here.  For example
+    - pytest>=3.9
```
navarroc, line 65: Should this be 3.8?

ywkim312: is 3.8 okay? If so, I will modify it

navarroc: I'm just trying to understand why we set 3.9 here. Have we been using 3.9 in the conda recipe previously?

ywkim312: We've been using 3.6 so far

longshuicy:
leaving a comment here to test

recipes/meta.yaml
navarroc: Just going to leave this year, we realized this is the pytest version not python version so this comment can be ignored.

PR #29, opened 2021-10-05
Add Lisa Wang fixes.

nunezco2:
Added minor changes to the analysis indicated by Lisa to match the scientific specification of her analysis and obtain correct results.

Test data: [joplin_recovery.csv](https://github.com/IN-CORE/pyincore/files/7288668/joplin_recovery.csv)


longshuicy (approved):
Approving. 

Just want to double check if those warnings have always been there?
image

Issue #28, opened 2021-10-04
Add Joplin building functionality recovery time model


ncsa-mo:
Implement only the original, empirical part

PR #30, opened 2021-10-05
#28 add joplin building functionality recovery time model

ncsa-mo:
This is a working version of the empirical model
The analysis takes output of building damage, maps Damage states to Functionality levels (DS with the highest probability is the initial FL - initialFL). The target Functionality level (targetFL) is an input parameter and they said it will mostly be 0 - Full restoration. Output of the analysis is a csv file with restoration days.


diegoac2:
Can you please add the path of the new files to test_format.py?

Issue #32, opened 2021-10-13
remove calculate_custom_limit_state_w_conversion() and it's usage in epfdamage's liquefaction usage


vnarah2:
Also, test if EPF damage's liquefaction code works. If so, create a test script to test it.

More details on EPF change: https://github.com/IN-CORE/pyincore/pull/22#discussion_r721733621

PR #50, opened 2021-11-22
32 epfdamage liquefaction and LS name fixes. 

vnarah2:
See the #32 for more details of the task. It handles the below:
1. Fix the broken liquefaction logic on EPF Damage
2. Add test for liquefaction, remove "refactored" test because they are not needed any more
3. Remove deprecated LS/DS name specific method
4. Use 10 digit precision when adjusting PGD limit states

Use the test_epfddamage.py to test

### DIFF
--- a/pyincore/analyses/epfdamage/epfdamage.py
+++ b/pyincore/analyses/epfdamage/epfdamage.py
@@ -4,7 +4,6 @@
 # terms of the Mozilla Public License v2.0 which accompanies this distribution,
 # and is available at https://www.mozilla.org/en-US/MPL/2.0/
 
-import collections
 import concurrent.futures
 from itertools import repeat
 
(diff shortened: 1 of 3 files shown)


vnarah2:
@ncsa-mo we need to create an issue to update the documentation related to these spec changes (liquefaction_fragility_key addition etc.) . Here: https://incore-dev.ncsa.illinois.edu/doc/incore/analyses/epf_dmg.html

navarroc (changes requested):
Unused import on line 7, unused variables around line 57, extra space on line 290

navarroc:
Other than the few cosmetic things, this change looks good and I tested it with a notebook I had for epf_damage and verified with/without liquefaction works, setting the liquefaction key works as well.

navarroc:
We probably need to update the documentation to include the new liq_fragility_key.

longshuicy (approved):
LGTM approving

pyincore/analyses/epfdamage/epfdamage.py
```diff
@@ -205,87 +257,28 @@ def epf_damage_analysis_bulk_input(self, epfs, hazard_type, hazard_dataset_id,
...
+                    damage_result['liqdemandunits'] = None
+                    damage_result['liqhazval'] = None
+                    damage_result['liqprobability'] = None
 
             ds_results.append(ds_result)
             damage_results.append(damage_result)
 
             i += 1
 
         #############################################################
```
navarroc, line 281: Leftover comment line can probably be removed

navarroc (approved):
Besides the empty comment line, this looks good, approving

PR #56, opened 2021-12-02
55 add input dataset with target functionality levels

### LINKED ISSUE #55: add input dataset to joplin building functionality recovery time model
Add an optional dataset to the model with individual Target functionality levels (column). The idea is to set default target levels from the parameter (required) and use custom levels for subset of buildings (guid) provided in the input dataset (optional)

ncsa-mo:
close #55


navarroc:
This looks good. One quick question, should either the target_functionality_level parameter or the building_functionality_level dataset be required? If you set neither, the analysis will break, right? Maybe this is a question for the researchers, but I would think one of them should be required.

navarroc (approved):

ywkim312 (approved):

Issue #15, opened 2021-09-14
implement new get_building_period() method


diegoac2:
now period and num_stories is a part of expression existing in the fragility curve;

rewrote get_building_period() and need to implement that in various place

<p>dmg_probability = selected_fragility_set.calculate_limit_state_refactored_w_conversion(<br />hval_dict, **building_args, period=xxx)</p>
<p>&nbsp;</p>
<p>in building damage this period =xxx seems to be not needed, cuz it's already a part of the equation of the fragility curves</p>
<p>make sure i test removing it</p>
<p>&nbsp;</p>
<p>using the&nbsp;5b47b34e337d4a36290754a0 as example&nbsp;</p>
<div class="table-wrap">

description | "default building period"
-- | --
fullName | null
expression | "(0.097) * math.pow(num_stories * (13.0), 0.624)"

</div>
<p>&nbsp;</p>

PR #59, opened 2021-12-07
Implement new get building period() method #15

longshuicy:
TO test:

run test for `building damage`, `non-structural building damage` and `pipeline damage`

### DIFF
--- a/pyincore/analyses/buildingdamage/buildingdamage.py
+++ b/pyincore/analyses/buildingdamage/buildingdamage.py
@@ -171,7 +171,6 @@ class BuildingDamage(BaseAnalysis):
             num_stories = b['properties']['no_stories']
             selected_fragility_set = fragility_sets[b_id]
 
-            building_period = selected_fragility_set.fragility_curves[0].get_building_period(num_stories)
 
             # TODO: Once all fragilities are migrated to new format, we can remove this condition
             if isinstance(selected_fragility_set.fragility_curves[0], FragilityCurve):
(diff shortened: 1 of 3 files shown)


navarroc:
Can you point this at the release branch?

navarroc:
Tested this change and it looks good. I think we can remove the hack on line 458 though, right? We only hit that line because the fragilitycurve was always returning 0.0 for get_building_period.

navarroc (approved):

pyincore/analyses/buildingdamage/buildingdamage.py
```diff
@@ -171,7 +171,6 @@ def building_damage_analysis_bulk_input(self, buildings, retrofit_strategy, haza
             num_stories = b['properties']['no_stories']
```
navarroc, line 171: I think this is unused now

pyincore/analyses/buildingdamage/buildingdamage.py
```diff
@@ -171,7 +171,6 @@ def building_damage_analysis_bulk_input(self, buildings, retrofit_strategy, haza
...
             selected_fragility_set = fragility_sets[b_id]
```
navarroc, line 173: Extra blank line

navarroc:
analysisutil has an unused import.

longshuicy:
> analysisutil has an unused import.

removed

pyincore/models/fragilitycurve.py
```diff
@@ -101,31 +101,21 @@ def calculate_limit_state_probability(self, hazard_values: dict, fragility_curve
...
         return probability
 
-    def get_building_period(self, num_stories):
-        """Get building period from the fragility curve.
-        Args:
-            num_stories (int): Number of building stories.
-        Returns:
-            float: Building period.
-        """
+    def get_building_period(self, fragility_curve_parameters, **kwargs):
```
ncsa-mo, line 104: Can you add these docstings here:

def get_building_period(self, fragility_curve_parameters, **kwargs):
        """ Get building period from the fragility curve.
        
        Args:
            fragility_curve_parameters (dict): Fragility curve parameters.
            **kwargs: Keyword arguments.
        
        Returns:
            float: Building period.
            
        """

ncsa-mo (approved):

diegoac2 (approved):

Issue #69, opened 2022-01-05
update fragility specific functions so it deals with the service deprecations and using common DFR3Curve classes


vnarah2:
Fix the broken pytests and pyincore analyses that use fragilities.

Links to service change: IN-CORE/incore-services#48

PR #70, opened 2022-01-10
#69: Refactoring of DFR3 service to fix the broken fragility functions

vnarah2:
- Refactoring of DFR3 service to fix the broken fragility specific pytests and analyses 
- Repair changes are not included, it will be addressed in #71 
- All analyses and pytests should work now, except residentialbuildingrecovery that uses repair curves

Please ignore the inaccurate branch name (even though t says repair-curve-refactoring, this only deals with fragilities and broken repaircurve pytests)


longshuicy (approved):
Looks good. Approving

diegoac2 (approved):

PR #75, opened 2022-01-24
71 r3 refactoring to expressions

### LINKED ISSUE #71: update repair and restoration calculations to use expression based curves
Once the repair curves in database are migrated to new expression based format, we must update the calculation so it evaluates the expression instead of performing own calculation based on current alpha and beta values

Links to service change: https://github.com/IN-CORE/incore-services/pull/48

vnarah2:
Changes made:
- Added method to calculate the inverse of cdf which can be used in repair curve calculation to calculate repair time when percent of functionality is provided as an input. Used this in residentialbuildingrecovery analysis
- Changed the DFR3 expression evaluation method to also accept an array of numeric values, instead of a single value. This improves analyses performance where an array of x values can be evaluated together. 
- Added support for Restoration curves in pyincore. Included pytests.
- Updated the format of repair curves needed for residentialbuildingrecovery in DEV DB. Converted them to be stored as CDF instead of PPF

Things to test:
- See the new pytests included in this PR - repairs, restorations, evaluating curve expressions and its inverse
- Make sure residentialbuildingrecovery is not impacted due to repair curve change
- Make sure none of the fragility based analyses were impacted due to the changes made to `solve_curve_expression()`

Closes #71


tests/pyincore/models/test_dfr3curve.py
```diff
@@ -76,3 +94,45 @@ def test_calculate_limit_state_probability(fragility_set, hazard_values, args, e
...
+    elif type(result["PF_0"]) == numpy.float64:
+        assert result["PF_0"] == expected
+    else:
+        assert False
+
+
+@pytest.mark.parametrize("repair_set,args,expected", [
+    (get_repair_set("repairset.json"), {"repair_time": [0.5, 0.2]}, 2),
+    (get_repair_set("repairset.json"), {"repair_time": 0.67}, 27.50466741611462)
+])
```
longshuicy, line 115: The way you write the repair time expected is a little bit confused... 
2 is the length
 and 27.5xxx is the value
could you make them two different variables and evalute both len and value?

pyincore/models/dfr3curve.py
```diff
@@ -88,18 +89,50 @@ def calculate_limit_state_probability(self, hazard_values: dict, curve_parameter
...
+            any: Result of the evaluated inverse expression. Can be float, numpy.ndarray etc.
 
-            raise ValueError(error_msg)
+        """
+        inverse_rules = []
+        actual_rules = self.rules
+        for rule in self.rules:
+            if ".cdf(" in rule['expression']:
+                new_exp = rule['expression'].replace(".cdf(", ".ppf(")
+                inverse_rules.append({'condition': rule['condition'], 'expression': new_exp})
```
longshuicy, line 128: Keep repair time as x, and PF as y is very confusing. But i can't think of anything better at the moment. Could we discuss this issue in dev meeting? 
For now let's just keep it as it is.

tests/pyincore/models/test_dfr3curve.py
vnarah2: agreed. I can change it to an array value (instead of len) even for the list

pyincore/models/dfr3curve.py
vnarah2: you mean the other way? Keeping PF as x and time as y is confusing? If so, I agree, but don't have a better way to handle it at the moment. That's why I added a comment:
```
# NOTE: Even though the kwarg name is "repair_time", it actually takes  percent of functionality. DFR3
# system currently doesn't have a way to represent the name correctly when calculating the inverse.
```

pyincore/models/restorationcurveset.py
```diff
@@ -0,0 +1,121 @@
...
+
+        output = {}
+        index = 0
+
+        if len(self.restoration_curves) <= 5:
+            for restoration_curve in self.restoration_curves:
+                eval_value = restoration_curve.solve_curve_for_inverse(hazard_values={},
+                                                                       curve_parameters=self.curve_parameters, **kwargs)
+                output[restoration_curve.return_type['description']] = eval_value
+                index += 1
```
diegoac2, line 117: what is this variable doing?

longshuicy (approved):

diegoac2 (approved):

PR #83, opened 2022-01-28
78 pipeline restoration analysis

### LINKED ISSUE #78: water pipeline restoration model
see https://opensource.ncsa.illinois.edu/confluence/display/INCORE1/Water+Facility+and+Pipeline+restoration+analysis

vnarah2:
Closes #78 

This works but the repair times are very small for the input datasets being used, the reason being the break and leak rates are very small(all less than 1 day) in the pipeline damage output file (input of this analysis), and the diameter of the pipes being small (all have 2 inch diameter). I'm going to see if I can find other datasets/hazards that have higher damage and hence show larger repair times.

We should check:
1. if the smaller repair times in this case is ok and double check the expression being used. (should the repair rates be multiplied by 100 before calculating?)
2. are there other datasets (seaside?), that have more varied inputs than the shelby county one which has all pipes diameter set to 2 inches. If there is such dataset, I'll create another test
3. There is also a warning with this dataset, when it is loaded to a dataframe. It's probably not impacting anything, but worth checking `INFO - collection.py:crs() - Failed to auto identify EPSG: 7`

### DIFF
--- a/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
+++ b/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
@@ -188,11 +188,8 @@ class PipelineRestoration(BaseAnalysis):
 
         if 'guid' in dmg.keys():
             res_result['guid'] = dmg['guid']
-
-        elif 'id' in dmg.keys():
-            res_result['id'] = dmg['id']
         else:
-            res_result['id'] = 'NA'
+            res_result['guid'] = 'NA'
 
         res_result['breakrate'] = dmg['breakrate']
         res_result['leakrate'] = dmg['leakrate']
--- a/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
+++ b/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
@@ -191,9 +191,6 @@ class PipelineRestoration(BaseAnalysis):
         else:
             res_result['guid'] = 'NA'
 
-        res_result['breakrate'] = dmg['breakrate']
-        res_result['leakrate'] = dmg['leakrate']
-
         res_result['repair_time'] = restoration_set.calculate_restoration_rates(**{
             "break_rate": float(dmg['breakrate']),
             "leak_rate": float(dmg['leakrate']),
--- a/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
+++ b/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
@@ -99,7 +99,6 @@ class PipelineRestoration(BaseAnalysis):
         damage_result = pipelines_dmg_df.merge(pipelines_df, on='guid')
         damage_result = damage_result.to_dict(orient='records')
 
-        # setting number of cpus to use
         user_defined_cpu = 1
         if not self.get_parameter("num_cpu") is None and self.get_parameter(
                 "num_cpu") > 0:
@@ -167,11 +166,9 @@ class PipelineRestoration(BaseAnalysis):
         restoration_sets = self.restorationsvc.match_list_of_dicts(self.get_input_dataset("dfr3_mapping_set"),
                                                                    damage, restoration_key)
 
-        i = 0
         for dmg in damage:
             res = self.restoration_time(dmg, num_available_workers, restoration_sets[dmg['guid']])
             restoration_results.append(res)
-            i += 1
 
         return restoration_results
 
@@ -199,14 +196,11 @@ class PipelineRestoration(BaseAnalysis):
 
         res_result['breakrate'] = dmg['breakrate']
         res_result['leakrate'] = dmg['leakrate']
-        res_result['diameter'] = dmg['diameter']
 
         res_result['repair_time'] = restoration_set.calculate_restoration_rates(**{
             "break_rate": float(dmg['breakrate']),
             "leak_rate": float(dmg['leakrate']),
             "pipe_length": dmg['length'],
             "num_workers": num_available_workers})['RT']
-
-        res_result['haz_expose'] = dmg['haz_expose']
 
         return res_result
--- a/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
+++ b/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
@@ -202,8 +202,8 @@ class PipelineRestoration(BaseAnalysis):
         res_result['diameter'] = dmg['diameter']
 
         res_result['repair_time'] = restoration_set.calculate_restoration_rates(**{
-            "break_rate": float(dmg['breakrate'])*100,
-            "leak_rate": float(dmg['leakrate'])*100,
+            "break_rate": float(dmg['breakrate']),
+            "leak_rate": float(dmg['leakrate']),
             "pipe_length": dmg['length'],
             "num_workers": num_available_workers})['RT']
 
--- a/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
+++ b/pyincore/analyses/pipelinerestoration/pipelinerestoration.py
@@ -91,14 +91,13 @@ class PipelineRestoration(BaseAnalysis):
     def run(self):
         """Executes pipeline restoration analysis."""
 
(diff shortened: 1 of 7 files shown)


pyincore/analyses/pipelinerestoration/pipelinerestoration.py
```diff
@@ -0,0 +1,212 @@
...
+
+        pipelines_df = self.get_input_dataset("pipeline").get_dataframe_from_shapefile()
+
+        pipeline_dmg = self.get_input_dataset("pipeline_damage").get_csv_reader()
+        pipelines_dmg_df = pd.DataFrame(list(pipeline_dmg))
+
+        damage_result = pipelines_dmg_df.merge(pipelines_df, on='guid')
+        damage_result = damage_result.to_dict(orient='records')
+
+        # setting number of cpus to use
```
diegoac2, line 102: unnecessary comment

pyincore/analyses/pipelinerestoration/pipelinerestoration.py
```diff
@@ -0,0 +1,212 @@
...
+            restoration_key = "Restoration ID Code"
+
+        restoration_sets = self.restorationsvc.match_list_of_dicts(self.get_input_dataset("dfr3_mapping_set"),
+                                                                   damage, restoration_key)
+
+        i = 0
+        for dmg in damage:
+            res = self.restoration_time(dmg, num_available_workers, restoration_sets[dmg['guid']])
+            restoration_results.append(res)
+            i += 1
```
diegoac2, line 174: I think this is an unused variable

navarroc:
pipeline damage break_rate and leak_rates are in units of repairs / km, it looks like the analysis always converts to repairs/km. This needs to be accounted for in the restoration curves. I noticed the curves referred to in this mapping say "meters" so I'm not sure if any conversion is happening or not.  Also, in the mapping 61f35f09903e515036cee106, one of the rules doesn't look right, it has:

"AND":[
"double length GT 300.0",
"double length LE 0"
]

vnarah2:
> pipeline damage break_rate and leak_rates are in units of repairs / km, it looks like the analysis always converts to repairs/km. This needs to be accounted for in the restoration curves. I noticed the curves referred to in this mapping say "meters" so I'm not sure if any conversion is happening or not. Also, in the mapping 61f35f09903e515036cee106, one of the rules doesn't look right, it has:
> 
> "AND":[ "double length GT 300.0", "double length LE 0" ]

@navarroc great observation on /km and pipelength being used is in meters. May be that explains why the repair time being calculated is too small. Let's discuss this in one of our meetings.

About the other condition, it is just there to catch the other cases(as else condition) which was described as default (class "U") for pipes. Table 8.1.c here: https://opensource.ncsa.illinois.edu/confluence/display/INCORE1/Water+Facility+and+Pipeline+restoration+analysis
I think the rule is ok, but could be improved to also consider null values in the dataset. I'm glad you brought this up, all the rules should actually be around "diameter", not "length". I'll fix it on the mappings

navarroc:
Shouldn't the catch all restoration curve be LT 8 or GT 300 to catch the 2" diameter pipes? 

> > pipeline damage break_rate and leak_rates are in units of repairs / km, it looks like the analysis always converts to repairs/km. This needs to be accounted for in the restoration curves. I noticed the curves referred to in this mapping say "meters" so I'm not sure if any conversion is happening or not. Also, in the mapping 61f35f09903e515036cee106, one of the rules doesn't look right, it has:
> > "AND":[ "double length GT 300.0", "double length LE 0" ]
> 
> @navarroc great observation on /km and pipelength being used is in meters. May be that explains why the repair time being calculated is too small. Let's discuss this in one of our meetings.
> 
> About the other condition, it is just there to catch the other cases(as else condition) which was described as default (class "U") for pipes. Table 8.1.c here: https://opensource.ncsa.illinois.edu/confluence/display/INCORE1/Water+Facility+and+Pipeline+restoration+analysis I think the rule is ok, but could be improved to also consider null values in the dataset. I'm glad you brought this up, all the rules should actually be around "diameter", not "length". I'll fix it on the mappings

vnarah2:
> Shouldn't the catch all restoration curve be LT 8 or GT 300 to catch the 2" diameter pipes?

Even though the result won't change, I agree that will make the rules match what's in the table. Made the change, please review again @navarroc

tests/pyincore/analyses/pipelinerestoration/test_pipelinerestoration.py
```diff
@@ -0,0 +1,36 @@
...
+    # shelby county pipelines
+    pipeline_restoration.load_remote_input_dataset("pipeline", "5a284f28c7d30d13bc081d14")
+    pipeline_restoration.load_remote_input_dataset("pipeline_damage", "61f36023c53b3620b6b614c6")
+
+    # Load fragility mapping
+    fragility_service = FragilityService(client)
+    mapping_set = MappingSet(fragility_service.get_mapping("61f35f09903e515036cee106"))
+    pipeline_restoration.set_input_dataset('dfr3_mapping_set', mapping_set)
+
+    pipeline_restoration.set_parameter("result_name","pipeline_restoration_times")
```
longshuicy, line 25: PEP8 complains about missing a space between result_name and pipeline_restoration_times

longshuicy (approved):

pyincore/analyses/pipelinerestoration/pipelinerestoration.py
```diff
@@ -0,0 +1,212 @@
...
+        Returns:
+            dict: A dictionary with id/guid and restoration time, along with some inventory metadata
+        """
+        res_result = collections.OrderedDict()
+
+        if 'guid' in dmg.keys():
+            res_result['guid'] = dmg['guid']
+
+        elif 'id' in dmg.keys():
+            res_result['id'] = dmg['id']
```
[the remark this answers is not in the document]
navarroc: @vnarah2 do you know what this is? I think this is ready to merge once we remove the extra fields (diameter and haz_expose)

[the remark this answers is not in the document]
vnarah2: removed the elif - it's caused by bad copy-paste from montecarlo analysis. This should be good to merge now.

pyincore/analyses/pipelinerestoration/pipelinerestoration.py
```diff
@@ -0,0 +1,203 @@
...
+            dict: A dictionary with id/guid and restoration time, along with some inventory metadata
+        """
+        res_result = collections.OrderedDict()
+
+        if 'guid' in dmg.keys():
+            res_result['guid'] = dmg['guid']
+        else:
+            res_result['guid'] = 'NA'
+
+        res_result['breakrate'] = dmg['breakrate']
```
navarroc, line 194: I guess along the same lines as removing hazard_expose, do we actually need to store breakrate and leakrate since they are in the damage file? Storing these again would be inconsistent with what we've done previously. Thoughts @vnarah2 @longshuicy? I'm ok to address this later.

vnarah2: My inclination is to include more columns to make it more readable without the need to join with other inputs(dmg result, input inventory etc.). But, if we want to keep it consistent, I'm ok to remove. Please let me know

vnarah2: @navarroc thinking about this again, I can see how someone could misinterpret breakrate and leakrate as part of the outputs of this analysis, when they are not. I'm ok to remove them for consistency

navarroc: That's another concern, they might think it's a different leakrate/breakrate. It would probably be better to show in the example notebook (now or in the future) how to combine the dmg + restoration into a single table and visualize it rather than adding extra columns not produced by this analysis.

vnarah2: sound good, removed. Now it only has two columns: guid and repair_time

longshuicy: this looks good

diegoac2 (approved):

navarroc (approved):

Issue #52, opened 2021-11-30
Refactor damage ration datasets to use consisten column names and updated utility classes


navarroc:
Buildings and Roads use Mean Damage Factor, Bridges uses Best Mean Damage Ratio, we should use Mean Damage Factor for both and add a column that maps to our DS_ damage states so the values can be referenced by DS_ keys. 

We can also remove the type row (java.lang.String) since it's unnecessary.

PR #109, opened 2022-03-02
53: remove mean damage data types for which we don't have damage ratios

vnarah2:
This task shouldn't impact anything. 
Non-structural mean damage is probably broken because the damage ratio datasets related to AS/DS are still using slight, moderate etc. instead of DS_X convention - This task doesn't fix it. It will be fixed in #52

### DIFF
--- a/pyincore/analyses/meandamage/meandamage.py
+++ b/pyincore/analyses/meandamage/meandamage.py
@@ -176,7 +176,7 @@ class MeanDamage(BaseAnalysis):
         """Calculates mean damage based on damage probabilities and damage ratios
 
         Args:
-            dmg (obj): dmg analysis output for a single building.
+            dmg (obj): dmg analysis output for a single entity in the built environment
             dmg_ratio_tbl (list): dmg ratio table.
             damage_interval_keys (list): damage interval keys
             is_bridge (bool): a boolean to indicate if the inventory type is bridge.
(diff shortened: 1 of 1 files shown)


navarroc:
NS damage I think works because we pass in the "keys" to pull the DS probabilities and in the meandamage code, we never assume row1 is Insignificant, we never actually parse the DS column from the damage ratios.

navarroc:
I just tested the mean_dmg.ipynb with the latest release and it works for non structural.

longshuicy (approved):

navarroc:
Can you fix the mean_damage method documentation? It says "Calculates building damgae results for a single building", this should be more generic, something like "Calculates mean damage based on damage probabilities and damage ratios"

vnarah2:
> Can you fix the mean_damage method documentation? It says "Calculates building damgae results for a single building", this should be more generic, something like "Calculates mean damage based on damage probabilities and damage ratios"

@navarroc Done

pyincore/analyses/meandamage/meandamage.py
```diff
@@ -173,10 +173,10 @@ def mean_damage_bulk_input(self, damage, dmg_ratio_tbl):
         return result
 
     def mean_damage(self, dmg, dmg_ratio_tbl, damage_interval_keys, is_bridge):
-        """Calculates building damage results for a single building.
+        """Calculates mean damage based on damage probabilities and damage ratios
 
         Args:
-            dmg (obj): dmg analysis output for a single entry.
+            dmg (obj): dmg analysis output for a single building.
```
navarroc, line 179: This is generic so it shouldn't say for a building. Maybe "damage analysis output for a single entity in the built environment"?

vnarah2: ah right! fixed. thanks for pointing it out

navarroc (approved):

Issue #113, opened 2022-03-09
index and improve the performance of restoration util


longshuicy:
- index guid
- add same approach to EPF

PR #114, opened 2022-03-09
Index and improve the performance of restoration util #113

longshuicy:
add index to guid
add same util to epf
shorten the name of epf

### DIFF
--- a/pyincore/analyses/epfrestoration/epfrestorationutil.py
+++ b/pyincore/analyses/epfrestoration/epfrestorationutil.py
@@ -1,0 +1,39 @@
+from pyincore.analyses.epfrestoration import EpfRestoration
+
+
+class EpfRestorationUtil:
+    def __init__(self, epfr: EpfRestoration):
+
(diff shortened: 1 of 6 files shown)


pyincore/analyses/epfrestoration/epfrestorationutil.py
```diff
@@ -0,0 +1,39 @@
+from pyincore.analyses.epfrestoration import EpfRestoration
+
+
+class EpfRestorationUtil:
+    def __init__(self, epfr: EpfRestoration):
```
navarroc, line 5: I wonder if a future improvement for this might be to allow users to provide the outputs from EpfRestoration directly to a utility method so it's available to users who have those datasets, but don't have the analysis object. I'm thinking for cases like DataWolf where the workflow output would be just the files from restoration, not the object. To use the utility in a downstream analysis, I'd have to reconstruct the EpfRestoration object. Same comment for the waterfacility utility.

navarroc (approved):

PR #121, opened 2022-03-23
93 add GitHub action to publish to conda

robkooper:
automatically push pyincore to conda


robkooper:
this code needs to be squashed merged

Issue #99, opened 2022-02-11
Housing recovery model analysis


ncsa-mo:
The housing recovery model using Census appraisal data, Galveston testbed notebook example. 
Notes: https://opensource.ncsa.illinois.edu/confluence/display/INCORE1/Housing+recovery+analysis

PR #134, opened 2022-04-22
Hotfix 1.4.1

nunezco2:
Fix HEAD issue on HousingRecovery

### DIFF
--- a/pyincore/analyses/housingrecovery/housingrecovery.py
+++ b/pyincore/analyses/housingrecovery/housingrecovery.py
@@ -176,17 +176,8 @@ class HousingRecovery(BaseAnalysis):
         hse_rec_fin2 = pd.DataFrame(hse_rec_fin_index, columns=index_yr)
         hse_recov = pd.concat([hse_recov, hse_rec_fin1, hse_rec_fin2], axis=1)
 
-<<<<<<< HEAD
         column_to_save = ["guid", "d_vacationct", "mhhinck", "pminoritybg", "dmg", "value_loss"] + bval_yr + index_yr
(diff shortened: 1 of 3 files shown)


pyincore/analyses/housingrecovery/housingrecovery.py
```diff
@@ -176,17 +176,8 @@ def run(self):
         hse_rec_fin2 = pd.DataFrame(hse_rec_fin_index, columns=index_yr)
         hse_recov = pd.concat([hse_recov, hse_rec_fin1, hse_rec_fin2], axis=1)
 
-<<<<<<< HEAD
         column_to_save = ["guid", "d_vacationct", "mhhinck", "pminoritybg", "dmg", "value_loss"] + bval_yr + index_yr
         hse_recov[column_to_save].to_csv(result_name + ".csv", index=False, float_format="%.2f")
-=======
-        if "dmg" in hse_recov.columns:
```
ywkim312, line 183: So this whole part doesn't need to be in the code, right?

nunezco2: Correct, based on Michal's #99 PR

ywkim312 (approved):

pyincore/analyses/housingrecovery/housingrecovery.py
navarroc: I don't see anywhere where we set the output, shouldn't there be a self.set_result_csv_data(...)? This is missing from the original branch #99 as well.

diegoac2 (approved):

navarroc (approved):

PR #157, opened 2022-06-23
Closes #155 - add earthquake liquefaction to building damage analysis

### LINKED ISSUE #155: Add earthquake liquefaction to building damage analysis
Add earthquake liquefaction to building damage analysis. To do this,

1. Update analysis inputs (requires liquefaction susceptibility)
2. Call liquefaction endpoint to get probability of ground failure
3. Update damage output to include liquefaction

navarroc:
Added the missing pieces to support building damage with liquefaction. This is based on the Ergo liquefaction implementation from Hazus. I followed much of the same process as pipeline damage repair rate in terms of variable names. The main difference is how liquefaction is applied. Pipelines have a separate fragility curve whereas buildings adjust the limit states using the ground failure probabilities. 

I updated the test_buildingdamage.py to include liquefaction. You can run with and without liquefaction to see the impact.

### DIFF
--- a/pyincore/utils/analysisutil.py
+++ b/pyincore/utils/analysisutil.py
@@ -223,26 +223,46 @@ class AnalysisUtil:
         return result_dataset_id
 
     @staticmethod
-    def adjust_limit_states_for_liquefaction(limit_state_prob, ground_failure_prob):
-        """Adjust limit state probabilities for liquefaction using Hazus methodology.
+    def adjust_damage_for_liquefaction(limit_state_probabilities, ground_failure_probabilities):
+        """Adjusts building damage probability based on liquefaction ground failure probability
+        with the liq_dmg, we know that it is 3 values, the first two are the same.
+        The 3rd might be different.
+        We always want to apply the first two to all damage states except the highest.
 
-            Args:
-                limit_state_prob: limit state probabilities
-                ground_failure_prob: probability of ground failure
+        Args:
+            limit_state_probabilities (obj): Limit state probabilities.
+            ground_failure_probabilities (list): Ground failure probabilities.
 
-            Returns:
-                dict: limit state probabilities adjusted to include liquefaction
+        Returns:
+            OrderedDict: Adjusted limit state probability.
 
         """
-        # Assumes 3 Limit states, 4 Damage States
-        limit_state_prob['LS_0'] = limit_state_prob.get('LS_0') + ground_failure_prob[0] - limit_state_prob.get(
-            'LS_0') * ground_failure_prob[0]
-        limit_state_prob['LS_1'] = limit_state_prob.get('LS_1') + ground_failure_prob[1] - limit_state_prob.get(
-            'LS_1') * ground_failure_prob[1]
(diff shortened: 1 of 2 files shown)


nunezco2 (approved):

pyincore/utils/analysisutil.py
```diff
@@ -222,6 +222,28 @@ def create_result_dataset(datasvc: DataService, parentid: str,
...
+        # Assumes 3 Limit states, 4 Damage States
+        limit_state_prob['LS_0'] = limit_state_prob.get('LS_0') + ground_failure_prob[0] - limit_state_prob.get(
+            'LS_0') * ground_failure_prob[0]
+        limit_state_prob['LS_1'] = limit_state_prob.get('LS_1') + ground_failure_prob[1] - limit_state_prob.get(
+            'LS_1') * ground_failure_prob[1]
+        limit_state_prob['LS_2'] = limit_state_prob.get('LS_2') + ground_failure_prob[2] - limit_state_prob.get(
+            'LS_2') * ground_failure_prob[2]
+
+        return AnalysisUtil.update_precision_of_dicts(limit_state_prob)
+
```
longshuicy, line 246: There is the similar method in `nonstructurebuildingdamge`
```
@staticmethod
    def adjust_damage_for_liquefaction(limit_state_probabilities, ground_failure_probabilities):
        """Adjusts building damage probability based on liquefaction ground failure probability
        with the liq_dmg, we know that it is 3 values, the first two are the same.
        The 3rd might be different.
        We always want to apply the first two to all damage states except the highest.

        Args:
            limit_state_probabilities (obj): Limit state probabilities.
            ground_failure_probabilities (list): Ground failure probabilities.

        Returns:
            OrderedDict: Adjusted limit state probability.

        """
        keys = list(limit_state_probabilities.keys())
        adjusted_limit_state_probabilities = collections.OrderedDict()

        for i in range(len(keys)):
            # check and see...if we are trying to use the last ground failure
            # number for something other than the
            # last limit-state-probability, then we should use the
            # second-to-last probability of ground failure instead.

            if i > len(ground_failure_probabilities) - 1:
                prob_ground_failure = ground_failure_probabilities[len(ground_failure_probabilities) - 2]
            else:
                prob_ground_failure = ground_failure_probabilities[i]

            adjusted_limit_state_probabilities[keys[i]] = \
                limit_state_probabilities[keys[i]] + prob_ground_failure \
                - limit_state_probabilities[keys[i]] * prob_ground_failure

        # the final one is the last of limitStates should match with the last of ground failures
        j = len(limit_state_probabilities) - 1
        prob_ground_failure = ground_failure_probabilities[-1]
        adjusted_limit_state_probabilities[keys[j]] = \
            limit_state_probabilities[keys[j]] \
            + prob_ground_failure - limit_state_probabilities[keys[j]] * prob_ground_failure

        return adjusted_limit_state_probabilities
```

navarroc: Good point, opened https://github.com/IN-CORE/pyincore/issues/159 to review these.

longshuicy (approved):
Approving. Looks good.

PR #158, opened 2022-06-23
pip, pypi and requirements

robkooper:
- auto push to pypi
- cleanup of requirements
-


nunezco2 (approved):

robkooper:
chatting with @jonglee1 , this will generate to restrictive requirements that will break installation of other packages.

.github/workflows/requirements.yaml
```diff
@@ -0,0 +1,37 @@
+name: pyclowder
```
navarroc, line 1: pyincore?

robkooper: same copy and paste errorr.

navarroc (approved):

recipes/meta.yaml
```diff
@@ -6,7 +6,7 @@ package:
   version: {{ version }}
 
 source:
-  path: /path/to/pyincore
+  path: /Users/kooper/git/incore/pyincore
```
ywkim312, line 9: can we change this to like ../ ?

nunezco2 (approved):

PR #171, opened 2022-07-06
WIP 168 replace social vulnerability computation with dataset coming from sv analysis

longshuicy:
Everything works except line 
```
 mv_index = np.where(tpm[:, 0] == sv_scores[household])[0][0]
```

Originally sv_scores are random values between lower and uppper bound (< 1.0) and tpm is also in that range; 
now the calculated sv_scores some times are > 1 which breaks the code.

Currently in discussion with Amin about this issue.


nunezco2 (approved):

ywkim312 (approved):
test run successfully. generated csv file. Approve

PR #172, opened 2022-07-08
Closes #169 - added discretized restoration output to epf

### LINKED ISSUE #169: epf restoration should output discretized functionality based on damage
epf restoration should output discretized functionality based on damage. This will require the following changes:

1. The inventory dataset input should be replaced by the damage input
2. New output that gives discretized restoration for day 1, 3, 7, etc

The original analysis output will still be available for those who want to combine that output with other approaches to functionality.

navarroc:
As discussed in the last dev meeting, I updated EpfRestoration to output discretized restoration. This required two new inputs:

1) EPF Damage to provide the damage probabilities
2) A new input dataset - incore:epfDiscretizedRestoration - from Hazus table 8-28 (ingested as 62c886ae861e370172c71194)

The unit test is updated to chain epf damage and use the new input referenced above.

### DIFF
--- a/pyincore/analyses/epfrestoration/epfrestoration.py
+++ b/pyincore/analyses/epfrestoration/epfrestoration.py
@@ -9,7 +9,7 @@ Electric Power Facility Restoration
 
 import numpy as np
 
-from pyincore import BaseAnalysis, RestorationService
+from pyincore import BaseAnalysis, RestorationService, AnalysisUtil
 from pyincore.models.restorationcurveset import RestorationCurveSet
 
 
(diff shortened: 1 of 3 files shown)


ywkim312:
Can you update the Changelog? In pyincore-viz, we just add changelogs since there are not many issues.

nunezco2 (approved):

pyincore/analyses/epfrestoration/epfrestoration.py
```diff
@@ -169,7 +211,27 @@ def get_spec(self):
                     'required': True,
                     'description': 'DFR3 Mapping Set Object',
                     'type': ['incore:dfr3MappingSet'],
+                },
+                {
+
+                    'id': 'restoration_table',
```
longshuicy, line 217: Do we always assume we will have a discrete restoration table exist? I.e. is this analysis always going to calculate both continuous and discrete?

navarroc: That's a good question. We can discuss it in today's dev meeting or merge this and update this later.

PR #181, opened 2022-07-19
fix requirements runner

robkooper:
- added requests library to requirements.min
- update requirements document

error in runs was due to the fact that the `if` statement only works in steps.


.github/workflows/requirements.yaml
```diff
@@ -17,7 +17,6 @@ env:
 jobs:
   pyclowder:
```
navarroc, line 18: pyincore?

longshuicy (approved):

PR #187, opened 2022-07-21
Simplified EPF restoration and WF restoration to use continuous curves

navarroc:
I removed the table dataset input that was added as part of #169 and #170 for the EPF and WF restoration to get the values for specific days from the continuous curves. I updated the tests as well. There are minor differences in the results (1% or so) because I think the hazus tables were rounded to give whole numbers.


nunezco2 (approved):

longshuicy (approved):
I compared calculated EPF rate with the table. Looks good to me

PR #190, opened 2022-07-25
189 update fiona

ywkim312:
update fiona's minimum version


longshuicy (approved):

ywkim312:
There is a gdal error because of this update. Cancel the PR

PR #203, opened 2022-08-19
200 implement hurricane in galveston epf damage

longshuicy:


nunezco2 (approved):

ywkim312:
code looks okay to me. What is the way to test? Just simply run the pytest? Is there a notebook or something for testing this?

ywkim312 (approved):
tested with pytests test_epfdamage, test_populationdislocation and test_dfr3_service. All worked successfully.

tests/pyincore/analyses/populationdislocation/test_populationdislocation.py
```diff
@@ -14,7 +14,7 @@ def run_with_base_class():
     # incore-dev
     building_dmg = "602d96e4b1db9c28aeeebdce"  # dev Joplin
     # building_dmg = "602d975db1db9c28aeeebe35" # 15 guids test - dev Joplin
-    housing_unit_alloc = "5df7c989425e0b00092c5eb4"  # dev Joplin
+    housing_unit_alloc = "602ea965b1db9c28aeefa5d6"  # dev Joplin
```
navarroc, line 17: What happened to these? Were they deleted by the owner?

tests/pyincore/analyses/epfdamage/test_epfdamage.py
```diff
@@ -85,6 +85,25 @@ def run_with_base_class():
     # Run Analysis
     epf_dmg_tsu_re.run_analysis()
 
+    #############################################################################
+    # Galveston EPF damage
+    # Run epf damage
+    epf_dmg_eq_memphis = EpfDamage(client)
```
navarroc, line 91: Minor nitpick, the the variable has "memphis" in it, but the analysis is for galveston. Can we clean that up?

tests/pyincore/analyses/populationdislocation/test_populationdislocation.py
longshuicy: no idea.... i couldn't find it anywhere. Might due to what I did with dev mongodb last friday, but might not related at all since we don't run those python test for a while

navarroc: Interesting, it seems this has happened a few times lately. I'm not sure if it's external users deleting their own datasets (maybe to get under quota) and they inadvertently break tests. We may need to consider making a copy of some of these data for incore-tst user maybe.

navarroc:
This runs for me and looks good. I'd like to see my one comment addressed about variable names before this is merged, if possible.

longshuicy:
> tested with pytests test_epfdamage, test_populationdislocation, and test_dfr3service. Epfdamage and population dislocation was successful but dfr3service failed
> 
> Test_create_and_delete_fragility_set has an UnicodeDecodeError in line 141 It could be my local setting.... but not sure
> 
> ```
> test_dfr3service.py:141: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
> 
> fp = <_io.TextIOWrapper name='C:\\Workspace-incore2\\pyincore\\tests/data\\fragility_curve.json' mode='r' encoding='cp949'>
> cls = None, object_hook = None, parse_float = None, parse_int = None
> parse_constant = None, object_pairs_hook = None, kw = {}
> 
>     def load(fp, *, cls=None, object_hook=None, parse_float=None,
>             parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
>         """Deserialize ``fp`` (a ``.read()``-supporting file-like object containing
>         a JSON document) to a Python object.
>     
>         ``object_hook`` is an optional function that will be called with the
>         result of any object literal decode (a ``dict``). The return value of
>         ``object_hook`` will be used instead of the ``dict``. This feature
>         can be used to implement custom decoders (e.g. JSON-RPC class hinting).
>     
>         ``object_pairs_hook`` is an optional function that will be called with the
>         result of any object literal decoded with an ordered list of pairs.  The
>         return value of ``object_pairs_hook`` will be used instead of the ``dict``.
>         This feature can be used to implement custom decoders.  If ``object_hook``
>         is also defined, the ``object_pairs_hook`` takes priority.
>     
>         To use a custom ``JSONDecoder`` subclass, specify it with the ``cls``
>         kwarg; otherwise ``JSONDecoder`` is used.
>         """
> >       return loads(fp.read(),
>             cls=cls, object_hook=object_hook,
>             parse_float=parse_float, parse_int=parse_int,
>             parse_constant=parse_constant, object_pairs_hook=object_pairs_hook, **kw)
> E       UnicodeDecodeError: 'cp949' codec can't decode byte 0xe2 in position 2096: illegal multibyte sequence
> ```

this was not issue for me :-( 
The github test passed so im assuming we are ok?

PR #204, opened 2022-08-23
Combined building damage for wind, wave and surge

navarroc:
Combined building damage for wind, wave and surge

### DIFF
--- a/docs/source/modules.rst
+++ b/docs/source/modules.rst
@@ -41,6 +41,10 @@ analyses/capitalshocks
 ..  autoclass:: pyincore.analyses.capitalshocks.CapitalShocks
     :members:
 
+analyses/combinedwindwavesurgebuildingdamage
+..  autoclass:: pyincore.analyses.combinedwindwavesurgebuildingdamage.CombinedWindWaveSurgeBuildingDamage
+    :members:
+
 analyses/cumulativebuildingdamage
 =================================
 ..  autoclass:: cumulativebuildingdamage.cumulativebuildingdamage.CumulativeBuildingDamage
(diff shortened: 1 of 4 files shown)


longshuicy:
code runs and looks good to me. Two things i noticed:

- I saw a deprecated warning
```
/Users/cwang138/Documents/INCORE-2.0/pyincore/pyincore/analyses/combinedwindwavesurgebuildingdamage/combinedwindwavesurgebuildingdamage.py:51: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.
  max_val = combined_output[max_damage_states].max(axis=1)
```

- i notice the final output only documents the maximum state, not the maximum combined probability? Is that intended?

longshuicy (approved):

ywkim312 (approved):

navarroc:
> code runs and looks good to me. Two things i noticed:
> 
> * I saw a deprecated warning
> 
> ```
> /Users/cwang138/Documents/INCORE-2.0/pyincore/pyincore/analyses/combinedwindwavesurgebuildingdamage/combinedwindwavesurgebuildingdamage.py:51: FutureWarning: Passing a set as an indexer is deprecated and will raise in a future version. Use a list instead.
>   max_val = combined_output[max_damage_states].max(axis=1)
> ```
> 
> * i notice the final output only documents the maximum state, not the maximum combined probability? Is that intended?

Yes, because it's not clear what the value should be. there are two cases to consider: 

1) Wind could have a max probability of 0.3 and nax state of DS_2 and Surge/wave could have a max probability of 0.2 and max state of DS_3 leading to an overall state of DS_3. 
2) Two hazards predict DS_3, which should be the max probability? In this case we could look at the max probability since both predicted DS_3 and put the highest value, but again, it seemed better to wait for input from them on how to finish this analysis rather than assuming.

PR #205, opened 2022-08-23
Test galveston bldg dmg

longshuicy:


navarroc:
We might not need this since it will duplicate my analysis test for combined wind, wave and surge. I haven't pushed it yet, but my combined wind, wave and surge analysis will run all 3 of these analyses and then combine the damages.

PR #223, opened 2022-10-21
222 integerate electric power availability model

longshuicy:
implementation of electric power availability model

Those are on incore-dev
- Fragility curve 6351716b297f7611014c2206
- Mapping: 635171e3297f7611014c2207
- Substations points:635176081f950c126bcbe296
- Building: 603010a4b1db9c28aef5319f
- City polygon: 635175db1f950c126bcbe249
- Flood: 5f4d02e99f43ee0dde768406


longshuicy:
Note to myself:
- need to add changelog
- need to include new package into setup.py

PR #224, opened 2022-10-25
new dataset changes

Rashmil-1999:
This draft PR introduces a new Dataset class with new read and write methods to hide the underlying mechanism of reading and writing datasets and modularizes the Input/Output Classes. I have left TODO's where we still need to discuss validation mechanisms.

### DIFF
--- a/pyincore/__init__.py
+++ b/pyincore/__init__.py
@@ -23,6 +23,7 @@ from pyincore.utils.analysisutil import AnalysisUtil
 from pyincore.utils.popdisloutputprocess import PopDislOutputProcess
 from pyincore.utils.cgeoutputprocess import CGEOutputProcess
 from pyincore.dataset import Dataset, InventoryDataset, DamageRatioDataset
+from pyincore.new_dataset import NewDataset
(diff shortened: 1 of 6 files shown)


nunezco2 (changes requested):
The code looks good, but we need a couple of things:

a. Use your code to implement [ExampleAnalysis](https://github.com/IN-CORE/pyincore/blob/develop/pyincore/analyses/example/exampleanalysis.py) in order to have a sense of how the code would be instantiated.

b. Pass the PEP8 verifier to make sure the formatting is adequate.

Rashmil-1999:
Yes, I was thinking of implementing it somewhere so we can understand what changes can be made. Just have one question.
Do I have to make changes in the BaseAnalysis?

nunezco2:
Apologies for the unclear explanation. I should have said "pass the PEP8 formatting test correctly".

Rashmil-1999:
Oh, okay so is there a command I can run to format it correctly? because my pycharm doesn't auto format for some reason. I tried using the pep8 config file.

PR #227, opened 2022-11-07
226 update bridge damage to include liquefaction

longshuicy:
To test:

run the `test_bridgedamage.py` and see the output of liquefaction

### DIFF
--- a/pyincore/analyses/bridgedamage/bridgedamage.py
+++ b/pyincore/analyses/bridgedamage/bridgedamage.py
@@ -123,10 +123,14 @@ class BridgeDamage(BaseAnalysis):
                 "use_liquefaction") is not None:
             use_liquefaction = self.get_parameter("use_liquefaction")
 
+        # Get geology dataset id containing liquefaction susceptibility
+        geology_dataset_id = self.get_parameter("liquefaction_geology_dataset_id")
+
         fragility_set = self.fragilitysvc.match_inventory(self.get_input_dataset("dfr3_mapping_set"), bridges,
                                                           fragility_key)
 
         values_payload = []
+        values_payload_liq = []  # for liquefaction, if used
         unmapped_bridges = []
         mapped_bridges = []
         for b in bridges:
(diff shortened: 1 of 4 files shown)


navarroc:
This is close, but LS_2 is wrong because I think the building implementation assumes 3 LS not 4 LS like we have with bridges. The ground failure probability returned from the service in the test case is:

```json
"groundFailureProb": [
            0.0670335192852054,
            0.0670335192852054,
            0.013406703857041081
        ]
```

This was done intentionally since Ergo/MAEViz had 3 LS. The first two would use the formula
LS_0 = LS_0 + 0.067 - LS_0 * 0.067
LS_1 = LS_1 + 0.067 - LS_1 * 0.067
And the last limit state would use the other ground failure value instead of 0.067. For the bridges case, we should do:
LS_0 = LS_0 + 0.067 - LS_0 * 0.067
LS_1 = LS_1 + 0.067 - LS_1 * 0.067
LS_2 = LS_2 + 0.067 - LS_2 * 0.067
LS_3 = LS_3 + 0.0134 - LS_3 * 0.0134

It looks like 0.0134 is getting applied to both LS_2 and LS_3 when I double checked the math.

pyincore/analyses/bridgedamage/bridgedamage.py
```diff
@@ -193,6 +210,12 @@ def bridge_damage_analysis_bulk_input(self, bridges, hazard_type,
                         selected_fragility_set.calculate_limit_state(hval_dict,
                                                                      inventory_type="bridge",
                                                                      **bridge_args)
+
+                    if use_liquefaction and geology_dataset_id is not None and liquefaction_resp is not None:
+                        ground_failure_prob = liquefaction_resp[i][BridgeUtil.GROUND_FAILURE_PROB]
```
navarroc, line 215: The service returns ground failure expecting 3 LS. The first two limit states use the same probability of ground failure and the last uses the second (you will notice the first two values of ground failure probability are the same). There are actually only 2 values computed for ground failure. One potential workaround would be to do the following after you fetch the liquefaction response on line 215. Add:

# Adjust for 4 LS of bridges - the service returns ground failure expecting 3 LS
ground_failure_prob.append(ground_failure_prob[2])
ground_failure_prob[2] = ground_failure_prob[1]

This will add another ground failure probability to the list and then update the one at index 2 to match what is at 0 and 1. Other suggestions are welcome, but this would be a small change so we don't have to adjust the utility method that handles liquefaction.

longshuicy: This is the easiest fix to me. Updated.

navarroc (approved):

nunezco2 (approved):

PR #262, opened 2023-02-02
Combine and reformat 3 demand type hurricane building damages

navarroc:
Updated combined building damage to output a new building damage CSV that takes the maximum from the 3 damages. It also still outputs the max DS for each of the 3 hazards and the overall max DS.


nunezco2 (approved):

longshuicy (approved):
Code looks good. Approved. 
Running it might take some time. Maybe wait till the run finish and merge. I'll let you know

PR #267, opened 2023-02-10
265 cge data type does not need to include testbed

longshuicy:


ywkim312:
This also should be tested using the csv files you sent it previously, right? I got the same error when I ran the pytest 
```Traceback (most recent call last):
  File "c:\python39\lib\genericpath.py", line 19, in exists
    os.stat(path)
TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
python-BaseException
```

longshuicy:
> I got this error when I ran pytest. This is the same error that I had before and I tested the pre generated csv files but it looks different from the one that I did before. Should I test it with ipopt installed?
> 
> ```
>   File "c:\python39\lib\genericpath.py", line 19, in exists
>     os.stat(path)
> TypeError: stat: path should be string, bytes, os.PathLike or integer, not NoneType
> python-BaseException
> ```

I might shared with you output using old input dataset; or there are randomness in the CGE so the result doesn't match.

ywkim312 (approved):

nunezco2 (approved):

navarroc (approved):

Issue #274, opened 2023-02-22
Building economic loss analysis breaks when no occupancy multiplier is provided


22dylan:
The buildingeconloss analysis breaks when no occupancy multiplier is provided. The [documentation](https://incore.ncsa.illinois.edu/doc/incore/analyses/building_loss.html) says that occupancy_multiplier is not required for this analysis

longshuicy:
This is the wrong branch. we already merge the correct one

PR #275, opened 2023-02-22
Updated CHANGELOG.md with fix to issue #274

22dylan:


navarroc:
I think we can close this PR, the original PR has the changelog updates.

navarroc:
Hmmm....it's odd. If I checkout your fork, I can see the changelog update; however, the original PR of the fork doesn't have these commits. We might have to manually add the CHANGELOG and look into how to properly add forked PRs back to our repository.

navarroc (approved):
Tested with my local test_buildingeconloss.py and the change looks to fix so the multiplier dataset is optional.

longshuicy (approved):

PR #277, opened 2023-02-24
Add method to space service to add a dataset by space name

ywkim312:


navarroc:
@longshuicy Do you think we should also add a convenience method add_to_space_by_name that handles calling the new get_space_by_name and then internally it handles calling add_dataset_to_space, which takes a space ID?

pyincore/spaceservice.py
```diff
@@ -73,6 +73,24 @@ def get_space_by_id(self, space_id: str):
...
+            obj: HTTP response with the returned space information.
+
+        """
+        r = self.client.get(self.base_space_url, params={"name": space_name})
+        if r.status_code == 200:
+            return r.json()
+        else:
+            # return empty list
+            # todo should this be some error message json instead of empty json?
+            return []
```
longshuicy, line 92: Looking at other methods, i think they don't check the code and just let the method throw errors

longshuicy:
> @longshuicy Do you think we should also add a convenience method add_to_space_by_name that handles calling the new get_space_by_name and then internally it handles calling add_dataset_to_space, which takes a space ID?

That would be very helpful!

navarroc:
@ywkim312 Can you add a method so that a user can directly add a dataset to a space by name? I'm thinking something like:

add_to_space_by_name(space_name, dataset_id)

It would just call your other method to get the space ID and then call our existing add_dataset_to_space.

pyincore/spaceservice.py
ywkim312: changed to throw an error instead of returning an empty list

ywkim312:
> get the space ID and then call our existing add_dataset_to_space.

@navarroc In this method, should the dataset id in the original owners space should be removed? Or just add to new space, eventually making the given dataset id in two different spaces?

ywkim312:
@navarroc add dataset by name has been added. I also created new methods for deleting dataset from the space as well.

navarroc:
> > get the space ID and then call our existing add_dataset_to_space.
> 
> @navarroc In this method, should the dataset id in the original owners space should be removed? Or just add to new space, eventually making the given dataset id in two different spaces?

We should just add it to the new space. My thought is that knowing the ID of a space is not straightforward so nobody is it's hard for users to use the current method, add_dataset_to_space because it requires the UUID of the space and the dataset UUID. 

So, my thought is that we add another helper method that lets a user instead ask it to add the dataset to a space by the name since these are unique. This would simply call your new method to get the space ID and then it would call the old method for the user to add it by the space ID. Does that make sense?

ywkim312:
> > > get the space ID and then call our existing add_dataset_to_space.
> > 
> > 
> > @navarroc In this method, should the dataset id in the original owners space should be removed? Or just add to new space, eventually making the given dataset id in two different spaces?
> 
> We should just add it to the new space. My thought is that knowing the ID of a space is not straightforward so nobody is it's hard for users to use the current method, add_dataset_to_space because it requires the UUID of the space and the dataset UUID.
> 
> So, my thought is that we add another helper method that lets a user instead ask it to add the dataset to a space by the name since these are unique. This would simply call your new method to get the space ID and then it would call the old method for the user to add it by the space ID. Does that make sense?

Yes, thank you for the answer

pyincore/spaceservice.py
```diff
@@ -91,6 +109,57 @@ def update_space(self, space_id: str, space_json):
...
+            obj: HTTP response with the updated space.
+
+        """
+        space_id = self.get_space_by_name(space_name)[0]
+
+        response = self.add_dataset_to_space(space_id["id"], dataset_id)
+
+        return response
+
+    def delete_from_space_by_name(self, space_name, dataset_id):
```
navarroc, line 129: I might suggest "remove" instead of "delete". This may confuse users.

pyincore/spaceservice.py
```diff
@@ -91,6 +109,57 @@ def update_space(self, space_id: str, space_json):
...
+            obj: HTTP response with the updated space.
+
+        """
+        space_id = self.get_space_by_name(space_name)[0]
+
+        response = self.delete_dataset_from_space(space_id["id"], dataset_id)
+
+        return response
+
+    def delete_dataset_from_space(self, space_id: str, dataset_id: str):
```
navarroc, line 146: Same here, remove instead of delete.

ywkim312: I have changed the word delete to remove in every occurrences in the PR's change including changelog

pyincore/spaceservice.py
```diff
@@ -91,6 +109,57 @@ def update_space(self, space_id: str, space_json):
         response = r.json()
         return response
 
+    def add_to_space_by_name(self, space_name, dataset_id):
```
navarroc, line 112: Can you add "str" as the types for space_name and dataset_id?

ywkim312: added, and added the other methods that are missing these

pyincore/spaceservice.py
ywkim312: udpated

navarroc (approved):

longshuicy (approved):

PR #290, opened 2023-03-10
Updated SLC CGE code format and handled infeasible solution case

navarroc:
I cleaned up some of the formatting issues - there are still more. The main fix is if the solution is infeasible, we don't write out the result datasets so it's clear the analysis failed. 

You can test this by running the test_saltlakecge.py with and without the bad shock file I uploaded to dev.

### DIFF
--- a/pyincore/analyses/saltlakecge/saltlakecge.py
+++ b/pyincore/analyses/saltlakecge/saltlakecge.py
@@ -1964,6 +1964,8 @@ class SaltLakeCGEModel(BaseAnalysis):
             self.set_result_csv_data("gross-income", gross_income, name="gross-income", source="dataframe", index=True)
             self.set_result_csv_data("household-count", household_count, name="household-count", source="dataframe",
                                      index=True)
+        else:
+            raise ValueError("Solution infeasible - no output to save")
 
     def get_spec(self):
         return {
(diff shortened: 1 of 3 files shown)


pyincore/analyses/saltlakecge/saltlakecge.py
```diff
@@ -1890,30 +1939,31 @@ def run_solver(cons_filename, temp_file_name="tmp.py"):
...
+
+            self.set_result_csv_data("domestic-supply", domestic_supply, name="domestic-supply", source="dataframe",
+                                     index=True)
+            self.set_result_csv_data("pre-disaster-factor-demand", pre_disaster_demand,
+                                     name="pre-disaster-factor-demand", source="dataframe", index=True)
+            self.set_result_csv_data("post-disaster-factor-demand", post_disaster_demand,
+                                     name="post-disaster-factor-demand", source="dataframe", index=True)
+            self.set_result_csv_data("gross-income", gross_income, name="gross-income", source="dataframe", index=True)
+            self.set_result_csv_data("household-count", household_count, name="household-count", source="dataframe",
+                                     index=True)
```
longshuicy, line 1966: is it worth it to raise an error here or where you evaluate the terminal condition? This way datawolf step might become "FAILED" and then frontend widget won't be clickable. 
```
else:
    raise ValueError("some message") 
```

jonglee1 (approved):

pyincore/analyses/saltlakecge/saltlakecge.py
navarroc: I will push this

longshuicy (approved):

PR #296, opened 2023-03-17
Timeout hotfix

Vismayak:


longshuicy:
I think you missed 2 functions. 

1. You need to expose timeout here as well as inside this function when it's actually calling the `download_dataset_blob`
```
def get_dataset_blob(self, dataset_id: str, join=None):
...
 local_filename = self.download_dataset_blob(cache_data_dir, dataset_id)
 ```
 
 2.
 ```
 def delete_dataset(self, dataset_id: str):
 ```

Vismayak:
@longshuicy All request changes have been done 😄

ywkim312:
how to test this?

Rashmil-1999 (approved):

PR #302, opened 2023-03-23
added flag for clean columns for joining table dataset with source dataset

ywkim312:
@longshuicy @jonglee1 @navarroc There was already an method for joining the table dataset with source dataset. I just added another variable to clean up the columns. If the new variable is true, it will remove all the columns from source dataset except 'geometry' and 'guid'. So eventually. the output geodataframe will only have the columns from table dataset with geometry. The user can easily convert this GeoDataFrame to shapefile or geopakage using geopands. 

I am not very sure how we will handle this join process at this moment. We probably need to discuss more how to do it. However, having this PR will be helpful in some cases


pyincore/utils/datasetutil.py
```diff
@@ -17,30 +17,38 @@
 
 class DatasetUtil:
     @staticmethod
-    def join_datasets(geodataset, tabledataset):
+    def join_datasets(geodataset, tabledataset, clean_attributes=True):
```
longshuicy, line 20: could we default to "false" to be backwards compatible?

longshuicy (changes requested):

pyincore/utils/datasetutil.py
ywkim312: changed to false

ywkim312:
> @ywkim312 Good to know that we already have a method for that. Im most curious about how fast it is? Could you have some benchmark for the SLC data I shared?

@longshuicy Sure, would you be able to provide the dataset id for SLC data and the table data that I can test?

pyincore/utils/datasetutil.py
```diff
@@ -17,30 +17,38 @@
...
+            gdf = gdf[['geometry','guid']]
+
+        # joining and indexing
         join_gdf = gdf.set_index("guid").join(df.set_index("guid"))
 
         return join_gdf
 
     @staticmethod
-    def join_table_dataset_with_source_dataset(dataset, client):
+    def join_table_dataset_with_source_dataset(dataset, client, clean_attributes=True):
```
longshuicy, line 45: So sorry I missed a spot, same here ...

ywkim312: no worries. updated

longshuicy (approved):

PR #321, opened 2023-05-01
310 pandas 200 doesnt work with some of the analysis

longshuicy:
To test run below
- tests/pyincore/analyses/ncifunctionality/test_ncifunctionality.py
- tests/pyincore/analyses/combinedwindwavesurgebuildingdamage/test_combinedwindwavesurgebuildingdamage.py
- tests/pyincore/analyses/multiobjectiveretrofitoptimization/test_multiobjectiveretrofitoptimization.py - optional since this runs for hours

Test backwards compatiblity
- [x] pandas 2.0.0
- [x] pandas 1.5.3
- [x] pandas 1.4.1
- [x] pandas 1.3.0
- [x] pandas 1.2.0
- [x] pandas 1.1.0
- [ ] pandas 1.0.0 - doesn't work anymore
AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:

- [ ] pandas 0.2.4

Things left to test:

- [x] Combined damage 
- [x] Combined loss
- [x] Multiobject -- runs forever 2+ hours but looks like it's passing all the pandas part
- [x] Transportation is depending on PR #320 

--------------------------
To test:
Run all the major tests in /test folders


tests/pyincore/utils/test_popdisloutputprocess.py
```diff
@@ -40,7 +40,7 @@ def run_convert_pd_json_chained(client):
     # incore-dev
     building_dmg = "602d96e4b1db9c28aeeebdce"  # dev Joplin
     # building_dmg = "602d975db1db9c28aeeebe35" # 15 guids test - dev Joplin
-    housing_unit_alloc = "6177089781ce0838fbacd6a8"  # dev Joplin 2ev3
+    housing_unit_alloc = "63eb0fc2695e0f57262f7650"  # dev Joplin 2ev3
```
robkooper, line 43: did you intend to change these?

tests/pyincore/analyses/populationdislocation/test_populationdislocation.py
```diff
@@ -14,7 +14,7 @@ def run_with_base_class():
     # incore-dev
     building_dmg = "602d96e4b1db9c28aeeebdce"  # dev Joplin
     # building_dmg = "602d975db1db9c28aeeebe35" # 15 guids test - dev Joplin
-    housing_unit_alloc = "602ea965b1db9c28aeefa5d6"  # dev Joplin
+    housing_unit_alloc = "63eb0fc2695e0f57262f7650"  # dev Joplin
```
robkooper, line 17: did you intend to change these?

recipes/meta.yaml
```diff
@@ -45,7 +45,7 @@ requirements:
     - geopandas>=0.6.1
     - matplotlib>=2.1.0
     - networkx>=2.2
-    - pandas>=0.24.1,<2
+    - pandas>=0.24.1
```
robkooper, line 48: now sure how to test but does it still build with pandas that is < 2?

robkooper (changes requested):

tests/pyincore/utils/test_popdisloutputprocess.py
longshuicy: yes the old id were outdated

tests/pyincore/analyses/populationdislocation/test_populationdislocation.py
longshuicy: Yes the old id was outdated

recipes/meta.yaml
Rashmil-1999: So should I test it as is? by running a new install of this pyincore branch? will it update the pandas version?

navarroc: To test this, I think you can do two things:
1) Create a conda environment for the 1.10.0 release (which has pandas < 2) to verify it still works with less than version 
2) Then you could update the conda environment to pandas 2 and verify it still works.

navarroc: Nevermind, you'd need to copy the new analyses over. If you can, you could use a pip virtualenv so you can install the updated analyses and test before and after upgrading pandas.

longshuicy: I verified that pandas>= 1.1.0 works and updated the recipes of publish accordingly.

robkooper: any reason not to make that the minimum version? That was released July 1, 2020

longshuicy: Lower than that won't work anymore @robkooper 
```
 pandas 1.0.0 - doesn't work anymore
AttributeError: module 'numpy' has no attribute 'bool'.
np.bool was a deprecated alias for the builtin bool. To avoid this error in existing code, use bool by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use np.bool_ here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
```

robkooper (commented):
LGTM did not test though

robkooper (approved):

Vismayak (approved):

PR #323, opened 2023-05-03
Function to handle float conversion in dict 

Vismayak:
Tackles cases of empty strings

### DIFF
--- a/pyincore/utils/analysisutil.py
+++ b/pyincore/utils/analysisutil.py
@@ -71,13 +71,15 @@ class AnalysisUtil:
 
     @staticmethod
     def dmg_string_dict_to_dmg_float_dict(dmg_dict: dict):
-        float_dmg_dict = {'guid': dmg_dict['guid']}
+        float_dmg_dict = {}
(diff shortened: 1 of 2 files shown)


pyincore/utils/analysisutil.py
```diff
@@ -68,35 +69,40 @@ def float_list_to_decimal(num_list: list):
     def float_dict_to_decimal(num_dict: dict):
         return {key: Decimal(str(num_dict[key])) for key in num_dict}
 
+    @staticmethod
+    def dmg_string_dict_to_dmg_float_dict(dmg_dict: dict):
```
longshuicy, line 73: The logic looks good but I would give it more generic naming. 
The main function is to turn empty string to nan so maybe name it something to reflect that?  
Feel free to disregard my comment though it's not super necessary :-)

Rashmil-1999 (approved):
There is a Conflict in the Changelog file could you please resolve it and then merge this?

longshuicy (approved):

PR #329, opened 2023-05-09
324 pyincore client should display an error message regarding quota

Rashmil-1999:


Rashmil-1999:
For the first 2 images, I saw that it has been wrapped in status check so I thought I shouldn't change that. But I think I might have missed the Dataservice file's screenshot that you sent. I will verify that. Also, should I change the previous 2? or its okay to keep it that way as it seems to handle failed requests?

Rashmil-1999:
Ahh, I see, it was getting too long above when I make the call so I did it in line 451. So it's handled. I can move the `.json()` above as well but it would be just fine... let me know if you want me to make that change.

longshuicy:
> For the first 2 images, I saw that it has been wrapped in status check so I thought I shouldn't change that. But I think I might have missed the Dataservice file's screenshot that you sent. I will verify that. Also, should I change the previous 2? or its okay to keep it that way as it seems to handle failed requests?

I recommend making them consistent so future developers don't get confused. Same with your second comment

longshuicy (approved):
image
 
 Works for me when i reduce my allocation

Vismayak (approved):

PR #342, opened 2023-05-14
Added python version workflow

ylyangtw:
create PR to test the workflow


ylyangtw:
python 3.8 & 3.9 pass the unit tests.
python 3.10 & 3.11 have conflicts with the dependencies in environment.yml.

The results are here: https://github.com/IN-CORE/pyincore/actions/runs/4970283959/jobs/8894061503
image

robkooper (changes requested):
Can you have this merged with pytests workflow, since that is almost the same.

Also if tests are failing for 3.10 and 3.11 can we make an issue for those, and comment them out for now since we want tests to pass.

ylyangtw:
> Can you have this merged with pytests workflow, since that is almost the same.
> 
> Also if tests are failing for 3.10 and 3.11 can we make an issue for those, and comment them out for now since we want tests to pass.

Good idea! Just merged them and created an issue for 3.10/3.11 https://github.com/IN-CORE/pyincore/issues/343

robkooper (approved):

PR #348, opened 2023-05-24
Updated check to see if string is float

Vismayak:


Rashmil-1999:
Could you also add a Changelog entry?

longshuicy (approved):

Rashmil-1999 (approved):

PR #344, opened 2023-05-16
339 electric power facility repair cost analysis

longshuicy:
#### How to test: 
- run test **tests/pyincore/analyses/epfrepaircost/test_epfrepaircost.py** as it is

~~[data.zip](https://github.com/IN-CORE/pyincore/files/11491951/data.zip)~~

~~- Once finalized I can upload those datasets to incore services~~

#### Sample output
[seaside_epf_repair_cost.csv](https://github.com/IN-CORE/pyincore/files/11491953/seaside_epf_repair_cost.csv)

PR #352, opened 2023-05-30
349 water facility repair cost analysis

longshuicy:
Similar as this merged PR https://github.com/IN-CORE/pyincore/pull/344

To test:
Run `tests/pyincore/analyses/waterfacilityrepaircost/test_waterfacilityrepaircost.py`

### DIFF
--- a/tests/pyincore/analyses/waterfacilityrepaircost/test_waterfacilityrepaircost.py
+++ b/tests/pyincore/analyses/waterfacilityrepaircost/test_waterfacilityrepaircost.py
@@ -1,0 +1,32 @@
+from pyincore import IncoreClient
+from pyincore.analyses.epfrepaircost import EpfRepairCost
+import pyincore.globals as pyglobals
+
+
+def run_with_base_class():
+    client = IncoreClient(pyglobals.INCORE_API_DEV_URL)
+    # client = IncoreClient()
+
+    epf_repair_cost = EpfRepairCost(client)
+
+    # Seaside EPF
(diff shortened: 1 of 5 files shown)


ylyangtw:
Ran the test successfully and got a csv result having the same format as the sample provided in PR 344.
image

ylyangtw (approved):

tests/pyincore/analyses/waterfacilityrepaircost/test_waterfacilityrepaircost.py
```diff
@@ -0,0 +1,30 @@
+from pyincore import IncoreClient, Dataset
```
navarroc, line 1: Unused dataset import

longshuicy:
> Overall this looks good. I think in the future we should consider if we can combine repair costs since the code is nearly identical.

Yes could we talk about this in the next dev meeting. These two analyses have identical code so combining will make sense

navarroc:
> > Overall this looks good. I think in the future we should consider if we can combine repair costs since the code is nearly identical.
> 
> Yes could we talk about this in the next dev meeting. These two analyses have identical code so combining will make sense

Sounds good. Maybe we can make a BaseRepairCostAnalysis that has the bulk of the code and each extends it.

longshuicy:
> Can you add epfrepaircost and this new analysis to the modules.rst in the docs folder?

good catch. Done.
i will do the same for pipeline repair cost

navarroc (approved):

PR #360, opened 2023-06-21
337 refactor building damage analysis to support multiple hazards

longshuicy:
I made a few hacks mainly around the demand types
1. Since there is no way to tell which demand type belongs what hazard, I have to hardcode a list of allowed hazard type for tsunami and earthquake
2. Need to record the mapping of original demand type vs adjusted (hazard service) returned hazard type. Original code just assume the **same length and order** (read the hval_dict[j] part)

E.g. 
- 1.04 Sec Sa need to be converted to 1.04 SA before sending to the services (payload); but then when it's returned it further becomes 1.0 SA
- But the original equation in the fragility curve is using 1.04 Sec SA
- Hence I invented a dictionary to record that

Not proud of that code I wrote but I can't think of better ways without fundamentally change the pyincore as well as services code (demand type data model)

Anyway it works; to test:
- run the test_buildingdamage.py and compare with old results to see if backward compatible
- check the `seaside_multihazard_` result


pyincore/utils/analysisutil.py
```diff
@@ -469,7 +469,7 @@ def get_hazard_demand_type(building, fragility_set, hazard_type):
         return hazard_demand_type
 
     @staticmethod
-    def get_hazard_demand_types(building, fragility_set, hazard_type):
+    def get_hazard_demand_types_units(building, fragility_set, hazard_type):
```
longshuicy, line 472: Read this function :-(

pyincore/hazardservice.py
```diff
@@ -1054,3 +1055,13 @@ def search_hurricanewf(self, text: str, skip: int = None, limit: int = None, tim
         r = self.client.get(url, params=payload, timeout=timeout, **kwargs)
 
         return self.return_http_response(r).json()
+
```
longshuicy, line 1058: @navarroc here i added the method meant to be replaced by real API calls once we have that endpoint

navarroc:
I'm not sure if it will be a problem with the conda packaging, but allowedDemands.json doesn't get copied to the virtualenv when I run python setup.py install.

navarroc:
The file on disk is also lower case alloweddemands.json, but the reference is allowedDemands.json.

pyincore/analyses/buildingdamage/buildingdamage.py
```diff
@@ -121,66 +122,88 @@ def building_damage_analysis_bulk_input(self, buildings, retrofit_strategy, haza
...
+                        value_liq = {
+                            "demands": [""],
+                            "units": [""],
+                            "loc": loc
+                        }
+                        values_payload_liq.append(value_liq)
+                else:
+                    unmapped_buildings.append(b)
+
+            if hazard_type == 'earthquake':
```
navarroc, line 175: I thought you were going to add another check here for earthquake+tsunami, but what you did by splitting the types and iterating over the types is probably more generic and better.

navarroc:
Is it possible to create a pyincore alpha release from this branch? It might be good to keep this open and let them test this and if there are issues, we can fix them within this branch. We could also test conda installing it first to verify all the pieces are there to use the updated code. Thoughts?

longshuicy:
> allowedDemands.json

fixed

ywkim312 (approved):

PR #370, opened 2023-07-13
367 investigate why multiple datasets being created after running pytest

ylyangtw:
For testing, you can run `test_hazardservice.py` or only its function `def test_create_and_delete_earthquake(hazardsvc)`

This PR also includes workaround for pytest. It passes the workflow and doesn't require testing.

### DIFF
--- a/.github/workflows/conda.yaml
+++ b/.github/workflows/conda.yaml
@@ -52,7 +52,7 @@ jobs:
       - name: Install miniconda
         uses: conda-incubator/setup-miniconda@v2
         with:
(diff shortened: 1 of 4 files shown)


Rashmil-1999 (approved):
The pytests ran successfully for me but I see that the Docker action fails... Could you please check that? I am approving it right now but please check that...

Rashmil-1999:
Also the pytest fails

Rashmil-1999:
Can you please re run it once again and investigate the Issue?

longshuicy (approved):

Issue #180, opened 2022-07-18
Generate specific testing capabilities for lists in `get_spec`


nunezco2:
Nested types are not captured by pyincore nicely. This will be solved next release.

PR #373, opened 2023-07-20
add semantic service module

Rashmil-1999:
This module will add 3 major functions to fetch data from the semantics endpoint.

*Note:* The pytests here will fail because the pytests run on the dev URL and won't be able to be tested until Ya-Lan's incore-services [#180](https://github.com/IN-CORE/incore-services/pull/180) therefore the PR should only be merged after her changes are integrated into the develop branch.

There are 3 functions to test:

- `get_all_semantic_types` which fetches all types. It also has the ability to save the response into a JSON file by passing `save_json=True`. The path to the file with the name can be specified, else, it will store in `.incore` cache directory by default
- `get_semantic_type_by_name` which will get a type by its name. The name has to be an exact match, otherwise, it will raise an HTTPError. It also supports the `save_json` parameter
- `search_semantic_type` will search for the query name provided. It doesn't support the same JSON as I thought it may contain other matches which might not be needed. 

To test, try the following:
```
from pyincore import SemanticService, IncoreClient

client = IncoreClient("https://incore-dev.ncsa.illinois.edu")
smtsvc = SemanticService(client)

smtsvc.get_all_semantic_types(save_json=True, json_path=<path to your destination with filename.json>)
smtsvc.get_all_semantic_types(hyperlink=True, save_json=True, json_path=<path to your destination with filename.json>)

smtsvc.get_semantic_type_by_name(type_name="incore:wildfireDamageRaster", save_json=True, json_path=<path to your destination with filename.json>)
# will raise HTTPError as there is no wildfireDamageRaster
smtsvc.get_semantic_type_by_name(type_name="wildfireDamageRaster", save_json=True, json_path=<path to your destination with filename.json>)

smtsvc.search_semantic_type(query="wildfireDamageRaster")
# should return empty list
smtsvc.search_semantic_type(query="sehfiuehworfhuiweahr")
```


pyincore/semanticservice.py
```diff
@@ -0,0 +1,178 @@
...
+        client (IncoreClient): Service authentication.
+
+    """
+
+    def __init__(self, client: IncoreClient):
+        self.client = client
+        self.base_url = urljoin(client.service_url, "semantics/api/types")
+
+    @staticmethod
+    def return_http_response(http_response):
```
longshuicy, line 35: Not related to this task, i notice this function has been repetitively exist for every service; anyway we can place them in a centralized place?

pyincore/semanticservice.py
```diff
@@ -0,0 +1,178 @@
...
+            "limit": limit,
+            "detail": detail,
+        }
+        response = self.client.get(
+            url, params=payload, timeout=timeout, **kwargs
+        )
+        data = self.return_http_response(response).json()
+
+        if save_json:
+            # if json_path is None: then save to the default location (incore cache folder)
```
longshuicy, line 108: Save to cache will be very hard for user to find. Why not just pass empty path so it will save at where user run this method?

pyincore/semanticservice.py
```diff
@@ -0,0 +1,178 @@
...
+        return data
+
+    def get_semantic_type_by_name(
+        self,
+        type_name: str,
+        save_json: bool = False,
+        json_path: str = None,
+        timeout: Tuple[int, int] = (30, 600),
+        **kwargs,
+    ) -> Union[list, dict]:
```
longshuicy, line 126: why the return is type list and dict? I thought it should be uniuqe?

pyincore/semanticservice.py
```diff
@@ -0,0 +1,178 @@
...
+            timeout (tuple): Timeout for the request in seconds.
+                            First value is the connection timeout, second value is the read timeout.
+            **kwargs: Arbitrary keyword arguments.
+
+        Returns:
+            list or dict: List or dictionary of semantic type.
+
+        """
+
+        url = f"{self.base_url}/{type_name}"
```
longshuicy, line 142: Could we use `url = urljoin(self.base_url, type_name)` instead?

pyincore/semanticservice.py
```diff
@@ -0,0 +1,178 @@
...
+            timeout (tuple): Timeout for the request in seconds.
+                            First value is the connection timeout, second value is the read timeout.
+            **kwargs: Arbitrary keyword arguments.
+
+        Returns:
+            list or dict: List or dictionary of semantic type.
+
+        """
+
+        url = self.base_url + "/search"
```
longshuicy, line 173: Could we use url = urljoin(self.base_url, "search") instead?

longshuicy (changes requested):

pyincore/semanticservice.py
Rashmil-1999: hahah, I was starting to think that too

Rashmil-1999: where do you suggest?

pyincore/semanticservice.py
Rashmil-1999: ahh, good point. I'll do that.

pyincore/semanticservice.py
Rashmil-1999: the return type could be either a list or a dict, which is why I had put a union but it seems like even if there is 1 response, it returns a list. I'll change it to just be a list

pyincore/semanticservice.py
Rashmil-1999: for some reason, if I did that it wasn't working as it gave not valid URL error and when I printed it out, it seemed fine

pyincore/semanticservice.py
Rashmil-1999: Screenshot 2023-07-26 at 4 14 04 PM
This is the error I get. As you can see in the very next line I print the urljoin output.... and it only contains the other second parameter... Idk it just omits the first part

longshuicy:
/Users/cwang138/opt/miniconda3/envs/incoreRC1.9.0/bin/python /Users/cwang138/Library/Application Support/JetBrains/Toolbox/apps/PyCharm-P/ch-0/231.8109.197/PyCharm.app/Contents/plugins/python/helpers/pycharm/_jb_pytest_runner.py --path /Users/cwang138/Documents/INCORE-2.0/pyincore/tests/pyincore/test_semanticservice.py 
Testing started at 4:22 PM ...
Launching pytest with arguments /Users/cwang138/Documents/INCORE-2.0/pyincore/tests/pyincore/test_semanticservice.py --no-header --no-summary -q in /Users/cwang138/Documents/INCORE-2.0/pyincore/tests/pyincore

Connection successful to IN-CORE services. pyIncore version detected: 1.11.0
Successfully initialized Incore client and services. Using https://incore-dev.ncsa.illinois.edu
============================= test session starts ==============================
collecting ... collected 3 items

test_semanticservice.py::test_get_all_semantic_types 
test_semanticservice.py::test_get_semantic_type_by_name 
test_semanticservice.py::test_search_semantic_types 

========================= 2 failed, 1 passed in 0.77s ==========================
FAILED              [ 33%]
tests/pyincore/test_semanticservice.py:24 (test_get_all_semantic_types)
1 != 1

Expected :1
Actual   :1
<Click to see difference>

semanticsvc = <pyincore.semanticservice.SemanticService object at 0x7f8db11b48e0>

    def test_get_all_semantic_types(semanticsvc):
        order = "asc"
        skip = 0
        limit = 10
    
        # get all semantic types without hyperlink
        hyperlink = False
        semantic_types = semanticsvc.get_all_semantic_types(
            hyperlink=hyperlink,
            order=order,
            skip=skip,
            limit=limit
        )
>       assert len(semantic_types) == limit and len(semantic_types[0].split("//")) < 1, "Should not have hyperlink"
E       AssertionError: Should not have hyperlink
E       assert (10 == 10 and 1 < 1)
E        +  where 10 = len(['ergo:bridgeDamage', 'ergo:bridgeDamageRatios', 'ergo:bridges', 'ergo:buildingASDamageRatios', 'ergo:buildingContentDamageRatios', 'ergo:buildingDSDamageRatios', ...])
E        +  and   1 = len(['ergo:bridgeDamage'])
E        +    where ['ergo:bridgeDamage'] = <built-in method split of str object at 0x7f8d90580ee0>('//')
E        +      where <built-in method split of str object at 0x7f8d90580ee0> = 'ergo:bridgeDamage'.split

test_semanticservice.py:38: AssertionError
FAILED           [ 66%]ERROR - client.py:return_http_response() - A HTTPError has occurred 
HTTP Status code: 404
Error Message: Cannot find the type incore:wildFireDamageRaster version latest !

tests/pyincore/test_semanticservice.py:50 (test_get_semantic_type_by_name)
semanticsvc = <pyincore.semanticservice.SemanticService object at 0x7f8db11b48e0>

    def test_get_semantic_type_by_name(semanticsvc):
        semantic_type_exists = "incore:wildFireDamageRaster"
        semantic_type_not_exists = "wildFireDamageRaster"
    
        # find semantic type by name which exists
>       semantic_types = semanticsvc.get_semantic_type_by_name(semantic_type_exists)

test_semanticservice.py:56: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../pyincore/semanticservice.py:143: in get_semantic_type_by_name
    response = self.client.get(url, timeout=timeout, **kwargs)
../../pyincore/client.py:317: in get
    return self.return_http_response(r)
../../pyincore/client.py:166: in return_http_response
    http_response.raise_for_status()
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <Response [404]>

    def raise_for_status(self):
        """Raises :class:`HTTPError`, if one occurred."""
    
        http_error_msg = ""
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode("utf-8")
            except UnicodeDecodeError:
                reason = self.reason.decode("iso-8859-1")
        else:
            reason = self.reason
    
        if 400 <= self.status_code < 500:
            http_error_msg = (
                f"{self.status_code} Client Error: {reason} for url: {self.url}"
            )
    
        elif 500 <= self.status_code < 600:
            http_error_msg = (
                f"{self.status_code} Server Error: {reason} for url: {self.url}"
            )
    
        if http_error_msg:
>           raise HTTPError(http_error_msg, response=self)
E           requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://incore-dev.ncsa.illinois.edu/semantics/api/types/incore:wildFireDamageRaster

../../../../../opt/miniconda3/envs/incoreRC1.9.0/lib/python3.8/site-packages/requests/models.py:1021: HTTPError
PASSED               [100%]
Process finished with exit code 1

pyincore/semanticservice.py
longshuicy: client.py? Just random thoughts i haven't think through it yet.

pyincore/semanticservice.py
longshuicy: have you looked why this sometimes return list (on the controller)? Shouldn't "name" be unique that it's always just gonna be one?

pyincore/semanticservice.py
longshuicy: ah maybe it's the colon? Could you look at things related to urlencoding?

Rashmil-1999:
I have resolved all pytests... it should work now...

pyincore/semanticservice.py
Rashmil-1999: Lets resolve this in a separate issue.

pyincore/semanticservice.py
```diff
@@ -0,0 +1,167 @@
...
+        return data
+
+    def get_semantic_type_by_name(
+        self,
+        type_name: str,
+        save_json: bool = False,
+        json_path: str = "",
+        timeout: Tuple[int, int] = (30, 600),
+        **kwargs,
+    ) -> list:
```
longshuicy, line 120: to dict maybe?

longshuicy (approved):

pyincore/semanticservice.py
Rashmil-1999: no, the response we get be it single or multi, it's always a list... which is why I just kept to list

longshuicy: you are right. 
But we should change. IMO It's equivlent to /dfr3/{id} or /dataset/{id}; returning a list is a bit confusing :-(

Rashmil-1999: Yep, I agree.

ylyangtw (approved):
Test works. Looks good.

PR #374, opened 2023-07-24
Fix mismatch spec type

longshuicy:
identify those mismatch while coming up with script to get data types for semantics


Rashmil-1999:
Do I need to test anything? It seems like you have just added `[]`

ylyangtw (approved):
Looks good~ approve~

ywkim312 (approved):

PR #396, opened 2023-08-22
393 nonstructural building damage hazard exposure column name is too long

longshuicy:
@navarroc i took your suggestion to aggregate the ds/as hazard exposure into just one haz_expose to minimize the changing needed. 

Test it by running the pytest as well as the **mean_dmg.ipynb** from `incore_docs`


ylyangtw (approved):
notebook and pytest run without problems. approve~

pyincore/analyses/nonstructbuildingdamage/nonstructbuildingdamage.py
```diff
@@ -301,6 +300,7 @@ def building_damage_analysis_bulk_input(self, buildings):
             damage_result['hazardtype'] = hazard_type
             damage_result['hazardvals_as'] = hazard_vals_as
             damage_result['hazardvals_ds'] = hazard_vals_ds
+            damage_result['haz_expose'] = haz_expose
```
navarroc, line 303: I can't test this since NFS is down, but I think haz_expose should be part of the building_result. That's where we put it for building damage.

longshuicy: ohh good catch! Will update that

navarroc (approved):

PR #405, opened 2023-09-08
404 support local hazard

longshuicy:
This is ready to be reviewed. This is a huge PR with the below features:

1. Create parent class **Hazard** and child classes **Earthquake**, **Tornado**, **Tsunami**, **Flood** and **Hurricane**
2. Represent hazard dataset in combine with existing Dataset class. Created **HazardDataset** and child classes accordingly
3. Backward compatibility of instantiate hazard from **remote service**
4. Create hazard **from local file**
5. **Attach local dataset** (raster and shapefile) to the locally created hazard
6. Read raster values; calculate tornado wind speed from shapefile. The response is made identical to the response from services
7. Lastly, update the **Bridge Damage** as an example to demonstrate how to integrate it with analysis

To Test:
1. Run pytest **tests/pyincore/models/test_units.py**. --> it covers almost all scenarios
2. Run Bridge Damage test e.g. **test_bridgedamage.py**, or **test_mmsa_bridgedamage.py** and compare the results with some previous. See attached zip file
[legacy.zip](https://github.com/IN-CORE/pyincore/files/12580845/legacy.zip)

3. Run local hurricane test **test_bridgedamage_w_local_hazard.py** and compare with galveston bridge damage result

To Dos:
1. Double check model based hazard; I might miss a few fields (but none of the fields are used anyway except id)
2. For EQ, need to add logic to find nearest period --> create new task
3. For EQ, need to add logic to convert pga to pgv, sd, sv, etc --> create new task 
5. Include more unit conversion especially for wind speed --> create new task


navarroc:
I wonder if we should put the model implementations into some sub-directories to organize things. Thinking longer term, at least for earthquake and maybe even hurricane, we could consider some local model implementations with locally implemented attenuation models that are not in the service and we probably don't want that all in the model folder.

pyincore/models/hazard/tornado.py
```diff
@@ -0,0 +1,170 @@
...
+        Returns:
+            obj: Hazard values.
+
+        """
+        if self.id and self.id != "" and hazard_service is not None:
+            return hazard_service.post_tornado_hazard_values(self.id, payload, **kwargs)
+        else:
+            return self.calculate_wind_speed_uniform_random_dist(payload)
+
+    def calculate_wind_speed_uniform_random_dist(self, payload):
```
navarroc, line 80: The default actually uses a Mersenne twister for random values. We'll need to see if python random has an implementation.

longshuicy: > Almost all module functions depend on the basic function [random()](https://docs.python.org/3/library/random.html#random.random), which generates a random float uniformly in the half-open range 0.0 <= X < 1.0. Python uses the Mersenne Twister as the core generator. It produces 53-bit precision floats and has a period of 2**19937-1. The underlying implementation in C is both fast and threadsafe. The Mersenne Twister is one of the most extensively tested random number generators in existence.

does that mean it's already default to be Mersenne twister? https://docs.python.org/3/library/random.html#:~:text=Python%20uses%20the%20Mersenne%20Twister,random%20number%20generators%20in%20existence.

longshuicy:
> I wonder if we should put the model implementations into some sub-directories to organize things. Thinking longer term, at least for earthquake and maybe even hurricane, we could consider some local model implementations with locally implemented attenuation models that are not in the service and we probably don't want that all in the model folder.

Yes I just move them to a **hazard** subfolder. Are you suggesting create subfolder for each hazard type underneath?

longshuicy:
@navarroc we discussed for 5 more minutes after you left. I'm adding the backward compatibility to still use "hazard_id" and "hazard_type" as a result of the discussion. See my latest push for that https://github.com/IN-CORE/pyincore/pull/405/commits/e78d4a8e1f73b1f4825b6956f171b2866eceb005

pyincore/models/hazard/tornado.py
navarroc: Yes, the default in TornadoParameters on the service is Mersenne unless the user specifies to use the uniform distribution.

navarroc:
> > I wonder if we should put the model implementations into some sub-directories to organize things. Thinking longer term, at least for earthquake and maybe even hurricane, we could consider some local model implementations with locally implemented attenuation models that are not in the service and we probably don't want that all in the model folder.
> 
> Yes I just move them to a **hazard** subfolder. Are you suggesting create subfolder for each hazard type underneath?

What you have now is probably fine. We can always create sub-folders per hazard later if we end up getting model implementations and it's cluttering the repository.

pyincore/models/hazard/tornado.py
navarroc: The default in the service is to use a uniform real distribution using an apache commons math library. The idea is that the wind is random inside an EF zone, bounded by the wind speed for the EF box (e.g. EF1 is between 86 MPH and 111 MPH). That means at any point in the EF box a location could experience the maximum. You can see it here: https://github.com/IN-CORE/incore-services/blob/develop/server/hazard-service/src/main/java/edu/illinois/ncsa/incore/service/hazard/models/tornado/utils/TornadoCalc.java#L151

I think we need to check for the windSpeedMethod in the tornadoParameters and if it's 1 (default) we should use the uniform real distribution. The other option is what you have now where the speed increases linearly across the EF boundary so depending where you are in the box will impact what wind speed gets computed. I think they prefer the randomness for wind speed since that's more typical.

navarroc:
> > I wonder if we should put the model implementations into some sub-directories to organize things. Thinking longer term, at least for earthquake and maybe even hurricane, we could consider some local model implementations with locally implemented attenuation models that are not in the service and we probably don't want that all in the model folder.
> 
> Yes I just move them to a **hazard** subfolder. Are you suggesting create subfolder for each hazard type underneath?

The hazard folder needs an __init__.py

pyincore/models/hazard/tornado.py
longshuicy: Thanks for the information and the code. I think i understand a little bit more now.  Yes I can try the random uniform.

pyincore/models/hazard/tornado.py
```diff
@@ -0,0 +1,185 @@
...
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.MAX_WIND_SPEED
+                        else:
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.EF_WIND_SPEED[ef_box + 1]
+
+                        # Either generate random wind speed
+                        if True:
+                            random.seed(self.SEED)
+                            raw_wind_speed = random.uniform(bottom_speed, top_speed)
```
navarroc, line 139: It looks like this method has both linear interpolation and the uniform random, should all that extra code get removed?

longshuicy: Ah I commit my experiments to this PR... will remove those shortly

longshuicy: Done. Could you double check one more time.. my brain is not super working....

pyincore/models/hazard/tornado.py
```diff
@@ -0,0 +1,177 @@
...
+
+                        if ef_box == 5:  # EF5
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.MAX_WIND_SPEED
+                        else:
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.EF_WIND_SPEED[ef_box + 1]
+
+                        # generate random wind speed
+                        random.seed(self.SEED)
```
navarroc, line 133: I think we need to match the service implementation here. A while back we decided to use the seed and get a random seed value based on what is set by the user. The default is -1 for seed, which indicates no seed is set and this method would handle that:

https://github.com/IN-CORE/incore-services/blob/develop/server/hazard-service/src/main/java/edu/illinois/ncsa/incore/service/hazard/models/tornado/utils/TornadoCalc.java#L73

If -1, then it will use the system time and randomize that. If a user sets a seed value when creating the dataset, it will use that.

pyincore/models/hazard/tornado.py
```diff
@@ -0,0 +1,177 @@
...
+    "thresholdUnit": "mph",
+    "spaces": [
+        "xxx"
+    ],
+    "date": "2020-08-14T16:22:32+0000",
+    "datasetId": "xxx"
+    """
+
+    def __init__(self, metadata, ef_rating_field="ef_rating", ef_wind_speed=(65, 86, 111, 136, 166, 200),
+                 max_wind_speed=250.0, seed=1234):
```
navarroc, line 32: I think seed should be -1 here and we need to check tornadoParameters object to see if a seed value was set.

pyincore/models/hazard/tornado.py
```diff
@@ -0,0 +1,199 @@
...
+
+                        if ef_box == 5:  # EF5
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.MAX_WIND_SPEED
+                        else:
+                            bottom_speed = self.EF_WIND_SPEED[ef_box]
+                            top_speed = self.EF_WIND_SPEED[ef_box + 1]
+
+                        # generate random wind speed
+                        random.seed(self.get_random_seed(location))
```
longshuicy, line 136: @navarroc here is the updated logic 
See line 170 `def get_random_seed(self, location)` as well

navarroc:
Minor nitpick, I think hazardDataset should be all lower case.

navarroc (approved):

PR #416, opened 2023-09-29
415 add support for local hazard in analyses

longshuicy:
Run the additional test and compare with legacy results

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Added
 - Add commercial recovery analysis [#395](https://github.com/IN-CORE/pyincore/issues/395)
-
+- Capability to support for local hazard [#404](https://github.com/IN-CORE/pyincore/issues/404)
+- Add support for local hazard with backward compatibility to analyses [#415](https://github.com/IN-CORE/pyincore/issues/415)
 
 ### Fixed
 - Aggregate hazard exposure column for non-structural building damage analysis to avoid column name cutoff and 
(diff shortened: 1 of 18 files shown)


navarroc (approved):

navarroc:
I've tested and compared the damage analyses, all look to be matching the previous results with the hazard object.

ylyangtw:
@navarroc How do you do the tests? I have tried pytest but it gets nothing..

ylyangtw:
Not sure whether this matters... I encounter this in test_tornadoepndamage
image
Files are generated successfully tho

navarroc:
> @navarroc How do you do the tests? I have tried pytest but it gets nothing..

The tests Chen added all have a main function so you can run them using python test_tornadoepndamage_w_hazard_obj.py for example and then run the other test file in the tornadepndamage folder to compare with the legacy results.

ylyangtw (approved):
Tested all the added analyses, all work.
Also compared all the first line of csv files, all are the same.
Changes look good.

longshuicy:
> Not sure whether this matters... I encounter this in test_tornadoepndamage image Files are generated successfully tho



> Not sure whether this matters... I encounter this in test_tornadoepndamage image Files are generated successfully tho

i created an issue will investigate later

navarroc:
There were two issues with the unit tests failing. One was the incrtest user looks like they were over quota. I fixed that, but in doing so, I "may" have accidentally deleted a dataset it referred to when I cleaned up hazard files in the incrtest space marked as "pytest". I thought they were all created tests. I tested in a local branch and if you update test_create_hurricane_from_remote to use ID "5f11e50cc6491311a814584c", all tests should pass.

PR #430, opened 2023-10-30
Release 1.14.0

longshuicy:


docs/source/conf.py
```diff
@@ -35,7 +35,7 @@
 # The short X.Y version
 version = '1.13'
```
ylyangtw, line 36: After we decide the version, this also needs change?

ywkim312 (approved):

PR #435, opened 2023-10-31
Update component versions for python 3.8 test

ywkim312:


ylyangtw (changes requested):
may need to modify `requirements.min` as it generates others

longshuicy:
duplicated as #442

PR #449, opened 2023-11-09
448 new hazard objects are not defined in modulesrst

longshuicy:
Not sure how i can test it though...


ylyangtw (approved):
change looks good. approving

PR #456, opened 2023-11-30
455 update get allowed demands method to return fixed information if offline

longshuicy:
This PR: 
- Use parameter "offline" to control client
- Add allowed demand information to `hazardservice.py`
- When offline mode detected, use fixed "allowed demand" information, instead of talking to service

To test:
- Run `tests/pyincore/analyses/buildingdamage/test_buildingdamage_offline.py`
- Note due to the randomness in tornado, 2 runs won't have the same exact result.


ylyangtw:
image
Hmm no item is collected... Did I test it correctly? I am using this command `python -m pytest tests/pyincore/analyses/buildingdamage/test_buildingdamage_offline.py`

tests/pyincore/analyses/buildingdamage/test_buildingdamage_offline.py
```diff
@@ -0,0 +1,75 @@
+import os
+
+from pyincore import IncoreClient, FragilityCurveSet, MappingSet, Tornado, GeoUtil, Dataset, Mapping, FragilityService
```
ylyangtw, line 3: GeoUitl is not used either, probably can be removed too

longshuicy:
> image Hmm no item is collected... Did I test it correctly? I am using this command `python -m pytest tests/pyincore/analyses/buildingdamage/test_buildingdamage_offline.py`

Ah all the test against analyses are not "pytest". you might just want to run them as a normal python script.

tests/pyincore/analyses/buildingdamage/test_buildingdamage_offline.py
longshuicy: Good catch. Removed :-)

pyincore/client.py
```diff
@@ -171,43 +171,53 @@ class IncoreClient(Client):
 
     """
```
navarroc, line 172: Can you add the new argument to the Args list?

navarroc:
Not for this PR, but I wonder if we need to check for offline mode so someone doesn't try to use an offline client to do operations like create an EQ.

longshuicy:
> Not for this PR, but I wonder if we need to check for offline mode so someone doesn't try to use an offline client to do operations like create an EQ.

Good catch. I will create a task.

pyincore/globals.py
```diff
@@ -44,3 +44,298 @@
 SCIP_PATH = shutil.which("scip")
 
 DAMAGE_PRECISION = 10
+
+DEFAULT_ALLOWED_DEMANDS = {
```
navarroc, line 48: globals doesn't feel like the right place for this. Would this be better in either a new "hazardutil" under "utils" or maybe even somewhere within the models/hazard? It could be part of the "hazard.py" maybe? Thoughts?

longshuicy: Yes agree. 
I tried to put it under Hazard class (hazard.py) but failed due to circular dependency. In that case, Hazard import HazardService, and HazardService import Hazard...
So I end up creating another class and put it under hazardservice.py, let me know if that works?

navarroc (approved):

longshuicy:
> What is the meaning of "Add allowed demand information to pyglobals" in the main description? I mean what is working with pyglobals. Also, the only hazardservice uses the offline, because this PR is only for "get allowed demands" method?

- "pyglobals..." is outdated and I will revise shortly. After discussion with Chris I moved the "demand types" to HazardService.py as another class
- The get_allowed_demands are the only places that still talks to services which prevent user doing completely "offline" analysis. Hence by creating a local version of those values, it makes offline to be possible

navarroc:
> What is the meaning of "Add allowed demand information to pyglobals" in the main description? I mean what is working with pyglobals. Also, the only hazardservice uses the offline, because this PR is only for "get allowed demands" method?

The name is what was originally planned (putting the demand types in globals), but after discussion, it seemed better closer to the hazard service since it impacted the hazard service. 

It was discovered that when running with local data, the hazard service still needed to talked to the service to get the allowed demand types. But if a user is offline, that is not available. This adds them to pyincore so they are available for offline mode so now a user should be able to run pyincore entirely offline.

ywkim312 (approved):

PR #473, opened 2024-01-02
build core-cge-ml module

Rashmil-1999:
In this PR I have added a new module called the Core CGE ML where the idea is to implement all repetitive actions related to CGE ML here. Since loads of models will come in, we will create a wrapper class for that testbed and keep model coefficients, base values, and the related files in that analysis folder. The core CGE ML will be imported, fed the data it needs, and spit out the outputs. That will be the goal of all future analysis implementations about CGE ML.

Possible modification I can see -
- [**Current**] Right now the implemented wrapper class will have to extract all the required data and feed it to the core class. The reason for doing this was to avoid possible future file structure changes that will lead to modifying the core again. Hence, I left the parsing part for analysis to implement.
- [**Improvement**] Maybe assume file structure to write general implementations and add an optional argument to accept a parsing function.

Note: Since we will be implementing the analyses so I don't know if this improvement is really needed or not. We could easily structure inputs as required by the Core CGE class.

let me know what your thoughts are and what changes can be made.


Rashmil-1999:
I have added the implementation of mmsa cge with a test_mmsacge.py file. all you need to do is to run this to test it. Main thing to look for is improvements in implementation or some modification.

pyincore/analyses/mmsacge/mmsacge.py
```diff
@@ -0,0 +1,169 @@
...
+        -------
+        Tuple[Dict[str, List[str]], np.ndarray, np.ndarray, Dict[str, np.ndarray]]
+            Returns a tuple containing the following:
+            1. sectors: Dictionary containing the sectors for each factor
+            2. base_cap_factors: List of numpy arrays containing the base capital for each factor
+            3. base_cap: Numpy array containing the base capital
+            4. model_coeffs: Dictionary containing the model coefficients for each factor
+        """
+        logger.info("Parsing input files...")
+        base_cap_factors, base_cap = self.parse_base_vals(os.path.join(pyglobals.PYINCORE_PACKAGE_HOME, "analyses", "mmsacge", "base_val_shelby_MMSA.xlsx"))
```
longshuicy, line 147: Does it have to be xlsx? I can see there are many sheets inside.

pyincore/analyses/mmsacge/mmsacge.py
```diff
@@ -0,0 +1,169 @@
...
+        # arrange the capital shocks in the same order as the sectors
+        shocks = []
+        for sector in self.sectors["ds"]:
+            shocks.append(
+                sector_shocks.loc[sector_shocks["sector"] == sector.upper()]["shock"]
+            )
+        capital_shocks = np.array(shocks).reshape(1, -1)
+        super().run_core_cge_ml(
+            self.base_cap, capital_shocks, self.model_coeffs, self.base_cap_factors[:1] # need to remove this slice for full release
+        )
```
longshuicy, line 168: Could we pull the " self.set_result_csv_data" out to each analysis. And allow user to append their own custom name to the filename?

longshuicy (commented):
Could you add an entry in the changelog?

longshuicy:
> * left the parsing part for analysis to implement.

Maybe for now you can put the parsing part in a util (instead of on the mmsa class) so in the future other testbed can still reuse them? Just a thought.

pyincore/analyses/mmsacge/mmsacge.py
Rashmil-1999: yeah, so I had a chat with Jong and he said we will make others to give us CSV :P let them do the work hahaha.... I will simplify the code and push it this week.

pyincore/analyses/mmsacge/mmsacge.py
Rashmil-1999: I don't think I need to pull out the functions, I can try and use the spec parameters and see if they are supplied, I will use them to set the name, otherwise, we can make the function return data frames and let them save it..

Rashmil-1999:
> > * left the parsing part for analysis to implement.
> 
> Maybe for now you can put the parsing part in a util (instead of on the mmsa class) so in the future other testbed can still reuse them? Just a thought.

yeah, if we can standardize the input files from the econ teams, we can shift it to the core class itself...

pyincore/analyses/core_cge_ml/corecgeml.py
```diff
@@ -0,0 +1,266 @@
...
+                "after": np.squeeze(factor_base_cap_after).tolist(),
+            }
+
+        constructed_outputs = self.construct_output(predictions)
+
+        if constructed_outputs.get("ds", None) is not None:
+            self.set_result_csv_data(
+                "domestic-supply",
+                pd.DataFrame(constructed_outputs["ds"]),
+                name=self.get_parameter("domestic_supply_fname")
```
longshuicy, line 224: Is it where @navarroc suggested you could fix a portion of the fname for datawolf to pick it up? Maybe try
```
name = self.get_parameter("domestic_supply_fname") + "-domestic-supply"
```

requirements.min
```diff
@@ -16,4 +16,5 @@ requests>=2.31.0
 rtree>=1.1.0
 scipy>=1.11.3
 shapely>=2.0.2
+openpyxl>=3.1.2
```
longshuicy, line 19: what's openyxl needed for? I thought you are not parsing excel sheet anymore?

longshuicy (changes requested):

pyincore/analyses/core_cge_ml/corecgeml.py
longshuicy: Ah nvm i saw you address this in the other SLC PR. Thanks!

requirements.min
Rashmil-1999: ahh yes, I forgot. I'll remove this.

Rashmil-1999: Addressed in SLC PR

navarroc:
To fix the unit tests, I uploaded new files. Can you replace the following?

660d9435ce705a7e547a664e replaces 642f66e5f27db6680103c4ad
660d95bece705a7e547a6654 replaces 642f6230f27db66801038964
660d97bfce705a7e547a6659 replaces 642f6495f27db66801039434

pyincore/analyses/core_cge_ml/corecgeml.py
```diff
@@ -0,0 +1,266 @@
+""" Core CGE ML """
```
navarroc, line 1: Needs copyright

pyincore/analyses/core_cge_ml/__init__.py
```diff
@@ -0,0 +1,7 @@
+# Copyright (c) 2019 University of Illinois and others. All rights reserved.
```
navarroc, line 1: wrong copyright date

navarroc:
I tested slc cge with this PR and results look good and match results I saved from the SLC CGE branch before it got merged into this branch. Just a few minor things with the copyright. I have not looked through the code exhaustively.

navarroc (approved):

ywkim312 (approved):

longshuicy (approved):
I ran the slc and connect its output to the playbook utility function (tests/pyincore/utils/test_cgecsvoutputjson.py) with updated region and sector name. All seems to work well. Approving.

PR #527, opened 2024-03-18
508 implement slc ml cge

Rashmil-1999:
This PR adds ML SLC CGE. I still have to remove 4 files based on what I get in the email reply.

You can test it by installing the current branch and running:
`python tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py`

### DIFF
--- a/pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
+++ b/pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
@@ -105,8 +105,8 @@ class MlEnabledCgeSlc(CoreCGEML):
         for sector in self.cap_shock_sectors:
             if sector.upper() not in [v.upper() for v in sector_shocks["sector"]]:
                 raise ValueError(
-                    f"Sector {sector} not found in the sector shocks file with\n {sector_shocks['sector']} sectors. 
-                    \nPlease make sure you have used the correct capital shocks"
+                    f"Sector {sector} not found in the sector shocks file with\n {sector_shocks['sector']} sectors.\n" + 
+                    "Please make sure you have used the correct capital shocks"
                 )
             shocks.append(
                 sector_shocks.loc[sector_shocks["sector"] == sector.upper()]["shock"]


navarroc:
We need to update the modules.rst in the docs folder so it picks up the module documentation (docs/source/)

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,197 @@
...
+        self.cap_shock_sectors = cap_shock_sectors
+        super(MlEnabledCgeSlc, self).__init__(incore_client, sectors, labor_groups=[f"L{gp}" for gp in range(1, 5)]) # 4 labor groups
+
+    def get_spec(self):
+        return {
+            'name': 'Salt-Lake-cge',
+            'description': 'CGE model for Salt Lake City.',
+            'input_parameters': [
+                {
+                    "id": "domestic_supply_fname",
```
navarroc, line 99: This seems overly complicated. Why not just follow the pattern from the other analyses and set the resultName and then the output will be something like:
result_name + household-count.csv. So if a user sets nothing, then the output is just household-count.csv and if they set "SLC" then it's SLC_household-count.csv? It seems like a lot of extra parameters to allow control over each filename.

Also, this poses a problem for running this is a workflow if the output file can be anything. The workflow uses wildcards to find output files so it's better if there is a fixed portion of the name. Then household count can be find by telling the workflow to look for a file called:

"*household-count.csv"

Rashmil-1999: I see, I thought we wanted to have control over individual file names, I can change it to the resultName if that is required.

navarroc: @longshuicy What are your thoughts? I think my preference is to match other analyses and add a setResultName that is then used to name each output. If we want to let users set the names, that's fine, but then we need to make sure there is a static part to the name for the reasons above with finding output files.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+from pyincore import IncoreClient, Dataset, globals as pyglobals
+from pyincore.analyses.mlenabledcgeslc import MlEnabledCgeSlc
+
+client = IncoreClient()
+
+mlcgeslc = MlEnabledCgeSlc(client)
+
+curr_dir = os.path.dirname(__file__)
```
navarroc, line 10: Before this is merged, we should push the sector_shocks data to incore-dev and read it as a dataset. We shouldn't push data into the source repository.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+from pyincore import IncoreClient, Dataset, globals as pyglobals
+from pyincore.analyses.mlenabledcgeslc import MlEnabledCgeSlc
+
+client = IncoreClient()
```
navarroc, line 6: This should call dev

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
Rashmil-1999: So, for our flow, we can just run the analysis as is which generates the original names as found in old CGE. But nevertheless, I have removed the feature allowing users to name each file, and just added result_name input param which is optional unlike other analyses. This input is a prefix to the original file names so the filenames will still have the original names before file extension.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
Rashmil-1999: Yes, I have addressed this.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
Rashmil-1999: Done.

Rashmil-1999:
@navarroc This modifies the base CGE a little. Do you want me to point this PR to the core?

navarroc:
> @navarroc This modifies the base CGE a little. Do you want me to point this PR to the core?

I think we probably need to anyways because of the way the ML CGE PRs are all branched from the core CGE branch. Even without this change I think we would need to do that because otherwise when the core got merged/squashed, the individual CGE branches would contain the commits that got squashed and cause a problem. Also, since there is already a release branch for pyincore, I think the core branch needs to point to it instead of develop

tests/pyincore/analyses/mlenabledcgeslc/sector_shocks.csv
navarroc, line 1: I think sector_shocks.csv can be deleted now, right?

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,169 @@
+from typing import Tuple, List, Dict
```
navarroc, line 1: unused import? Also, it needs a copyright header (see housingrecoverysequential.py) and the year should be the current year. You might also want to look at that class for an example documentation. I'm not sure what they want there, we may need to update that later with input from them.

tests/pyincore/analyses/mlenabledcgeslc/sector_shocks.csv
Rashmil-1999: ahh, yes, I forgot that.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
```diff
@@ -0,0 +1,16 @@
+import os
```
longshuicy, line 1: unused import

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
```diff
@@ -0,0 +1,16 @@
...
+
+from pyincore import IncoreClient, globals as pyglobals
+from pyincore.analyses.mlenabledcgeslc import MlEnabledCgeSlc
+
+client = IncoreClient(pyglobals.INCORE_API_DEV_URL)
+
+mlcgeslc = MlEnabledCgeSlc(client)
+
+sector_shocks = "65fdeed7e42f3b0da56c4eef"
```
longshuicy, line 10: Is it possible to chain it with sector shock? Or you are missing the sector_to_building file?

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
longshuicy: Sorry late to the game. 

> and just added result_name input param which is optional unlike other analyses. 

This sounds good to me.

longshuicy (approved):
Code runs and the output looks correct. 
I only browse the logic without looking into the detail too much. 
If anyone, including myself, finds they have extra time, it would be beneficial to take a closer look at the code. But for now it looks good.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
Rashmil-1999: I don't understand the question. We only need sector shocks for this. Capital Shocks analysis needs sector to building file.

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
Rashmil-1999: done

tests/pyincore/analyses/mlenabledcgeslc/test_mlcgeslc.py
Rashmil-1999: Ohh, yes it can be chained. I have done that in the Docs PR.

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,173 @@
...
+                {
+                    'id': 'household-count',
+                    'parent_type': '',
+                    'description': 'CSV file of household count',
+                    'type': 'incore:HouseholdCount'
+                }
+            ]
+        }
+
+    def run_analysis(self) -> None:
```
navarroc, line 148: Why do you overrride the base analysis run_analysis method? Why not just implement this all in the "run" method?

navarroc: I think you could replace what you have above with:

def run(self):

That way the base analysis which checks the dataset input types doesn't get overridden.

Rashmil-1999: Yes, I didn't realize that. I have changed it now.

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -145,7 +145,7 @@ def get_spec(self):
             ]
         }
 
-    def run_analysis(self) -> None:
+    def run(self) -> None:
```
navarroc, line 148: This doesn't need the "-> None":  If you see the base class, it should return "True" at the end.

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,173 @@
...
+        capital_shocks = np.array(shocks, dtype=np.float64).reshape(1, -1)
+        # logger.info(f"capital_shocks shape: {capital_shocks.shape}")
+        
+        super().run_core_cge_ml(
+            self.base_cap,
+            capital_shocks,
+            self.model_coeffs,
+            self.base_cap_factors,
+        )
+        logger.info(f"Running {self.model} model completed.")
```
navarroc, line 173: After this line it should return True

navarroc (changes requested):
Need to remove the -> None on the run method and return True at the end after it executes

navarroc:
Final comment - can you put the spec definition at the bottom of the class?

navarroc:
Almost final - there are some pep8 warnings, can you fix those?

Rashmil-1999:
> Almost final - there are some pep8 warnings, can you fix those?

I don't see any warnings though. Could you tell me what those warnings are?

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,178 @@
...
+from pyincore.utils import parse_files
+
+logger = pyglobals.LOGGER
+
+
+class MlEnabledCgeSlc(CoreCGEML):
+
+    model = "Machine Learning Enabled Computable General Equilibrium - Salt Lake City "
+
+    #### Coefficients files
```
navarroc, line 23: My pycharm complains about too many # for a comment block, PEP 8: E266 too many leading '#' for block comment

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -0,0 +1,178 @@
...
+        sector_shocks = pd.read_csv(
+            self.get_input_dataset("sector_shocks").get_file_path("csv")
+        )
+        # arrange the capital shocks in the same order as the sectors
+        shocks = []
+
+        for sector in self.cap_shock_sectors:
+            if sector.upper() not in [v.upper() for v in sector_shocks["sector"]]:
+                raise ValueError(
+                    f"Sector {sector} not found in the sector shocks file with\n {sector_shocks['sector']} sectors. \nPlease make sure you have used the correct capital shocks"
```
navarroc, line 108: Here too about the line length

Rashmil-1999: I think we would need to add precommit here as well so for now I have addresses these. We can later look into adding a consistent formatting across developers later.

pyincore/analyses/mlenabledcgeslc/mlcgeslc.py
```diff
@@ -105,7 +105,8 @@ def run(self) -> bool:
         for sector in self.cap_shock_sectors:
             if sector.upper() not in [v.upper() for v in sector_shocks["sector"]]:
                 raise ValueError(
-                    f"Sector {sector} not found in the sector shocks file with\n {sector_shocks['sector']} sectors. \nPlease make sure you have used the correct capital shocks"
+                    f"Sector {sector} not found in the sector shocks file with\n {sector_shocks['sector']} sectors. 
+                    \nPlease make sure you have used the correct capital shocks"
```
navarroc, line 109: There's an error here now. I think the line \n needs to start with f"

Rashmil-1999: I think now it should be good.

navarroc (approved):

PR #528, opened 2024-03-21
526 update the post processing step for generating population dislocation map layer

ylyangtw:
for testing, you can run `test_popdisloutputprocess.py` and look at the created file `joplin-pop-disl-numprec.shp`
you should see a lot more data with numprec=0 in the shape file

### DIFF
--- a/pyincore/utils/popdisloutputprocess.py
+++ b/pyincore/utils/popdisloutputprocess.py
@@ -46,30 +46,27 @@ class PopDislOutputProcess:
                 pd_result_flag = pd_result[(pd_result["guid"].notnull()) &
                                            (pd_result["numprec"].notnull()) &
                                            (pd_result["plcname10"] == filter_name)]
-                # only keep guid, place
+                # only keep guid and place
                 pd_result_shp = pd_result[(pd_result["guid"].notnull()) &
                                           (pd_result["numprec"].notnull()) &
                                           (pd_result["plcname10"] == filter_name)]
             else:
                 pd_result_flag = pd_result[(pd_result["guid"].notnull()) &
                                            (pd_result["numprec"].notnull())]
-                # only keep guid and dislocated
+                # only keep guid
                 pd_result_shp = pd_result[(pd_result["guid"].notnull()) &
                                           (pd_result["numprec"].notnull())]
         else:
             if filter_name:
                 pd_result_flag = pd_result[(pd_result["numprec"].notnull()) &
                                            (pd_result["plcname10"] == filter_name)]
-                # only keep guid, place
+                # only keep guid and place
                 pd_result_shp = pd_result[(pd_result["numprec"].notnull()) &
                                           (pd_result["plcname10"] == filter_name)]
             else:
                 pd_result_flag = pd_result[(pd_result["numprec"].notnull())]
                 # only keep guid
                 pd_result_shp = pd_result[(pd_result["numprec"].notnull())]
-
-        # Update "numprec" to 0 where "dislocated" is False
-        pd_result_shp.loc[pd_result_shp["dislocated"], "numprec"] = 0
 
         self.vacant_disl = vacant_disl
         self.pop_disl_result = pd_result_flag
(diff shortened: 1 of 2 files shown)


pyincore/utils/popdisloutputprocess.py
```diff
@@ -46,31 +46,36 @@ def __init__(self, pop_disl_result, pop_disl_result_path=None,
...
                                            (pd_result["plcname10"] == filter_name)]
-                # only keep guid, place and dislocated
-                pd_result_shp = pd_result[(pd_result["dislocated"]) &
-                                          (pd_result["guid"].notnull()) &
+                # only keep guid, place
+                pd_result_shp = pd_result[(pd_result["guid"].notnull()) &
                                           (pd_result["numprec"].notnull()) &
                                           (pd_result["plcname10"] == filter_name)]
+                # Update "numprec" to 0 where "dislocated" is False
+                pd_result_shp.loc[pd_result_shp["dislocated"] == False, "numprec"] = 0
```
longshuicy, line 54: I feel like you might be able to pull this ` pd_result_shp.loc[pd_result_shp["dislocated"] == False, "numprec"] = 0` outside all the if statement; because this should apply to everything regardless of the criteria?

longshuicy: == False --> 
Could you check the `pd_result_shp["dislocated"]` is it indeed a boolean?

longshuicy (changes requested):
I notice this numprec becomes float number, could it be integer?
before it was integer
image

image

pyincore/utils/popdisloutputprocess.py
ylyangtw: Thanks for catching this! pd_result_shp["dislocated"] is a boolean and shouldn't compare with False.

ylyangtw: moved `pd_result_shp.loc[pd_result_shp["dislocated"] == False, "numprec"] = 0` outside of all if statement, good suggestion!
set numprec as integer as well

navarroc:
Should this point at the release branch if we want to include this in the RC?

pyincore/utils/popdisloutputprocess.py
```diff
@@ -46,31 +46,31 @@ def __init__(self, pop_disl_result, pop_disl_result_path=None,
...
             else:
                 pd_result_flag = pd_result[(pd_result["numprec"].notnull())]
-                # only keep guid and dislocated
-                pd_result_shp = pd_result[(pd_result["dislocated"]) &
-                                          (pd_result["numprec"].notnull())]
+                # only keep guid
+                pd_result_shp = pd_result[(pd_result["numprec"].notnull())]
+
+        # Update "numprec" to 0 where "dislocated" is False
+        pd_result_shp.loc[pd_result_shp["dislocated"], "numprec"] = 0
```
navarroc, line 72: Correct me if I'm wrong @longshuicy, but I don't think this is necessary. I think numprec is just the number of people in the household so the fact they aren't dislocated doesn't need to modify anything. I think this step can be removed, right?

navarroc: Now I understand, the numprec value is getting populated with the value coming from HUA if the household dislocates. Since we are including all households, any that aren't dislocated should have a numprec of 0. Thanks @longshuicy

navarroc (approved):

pyincore/utils/popdisloutputprocess.py
longshuicy: @ylyangtw maybe just to be safe, could we just rename this field to "numprec_dislocated" so the original numprec still left untouched. Do you think it's worth it?

ylyangtw: If we rename it to "numprec_dislocated", the non-dislocated items under original numprec need to be False/0/Null. These two columns will be duplicated I think. So probably we don't need to keep the original.

longshuicy: Let's not do that then. I will approve your PR as it is :-)

longshuicy (approved):
Looks good now. approve.

PR #545, opened 2024-04-18
Fix permission error

ywkim312:
The thing happens when trying to delete the shapefile that involved in creating new dataset by using the following function

DatasetUtil.construct_updated_inventories

I suspect that the permission error might happen in just simply using the dataset by downloading from data service. However, in that case, it will just be saved in the cache directly and will not try to delete the file like this case. I have tried to close the dataset but it did not work because *.shp, and *.dbf files are being used by some windows process. So instead of trying to deleted it, I am just pushing the proper try and catch logic so the process go till the end instead of being interrupted and stopped in the middle of the process. 

Things that are tested in the PR before commit (nothing worked)

- close the dataset
- changed dataset to None
- tried to kill process that is connected to the file (high risk to make the system error)
- delete file in the folder one by one then delete the folder
- delete again after failure giving certain time of interval


pyincore/dataset.py
```diff
@@ -343,8 +344,12 @@ def delete_temp_folder(self):
         """
         path = Path(self.local_file_path)
         absolute_path = path.parent.absolute()
+
         if os.path.isdir(absolute_path):
-            shutil.rmtree(absolute_path)
+            try:
+                shutil.rmtree(absolute_path)
+            except PermissionError as e:
```
longshuicy, line 351: Since you are catch, would it be beneficial to catch more generic errors as well?

ywkim312: I haven't found any other things to catch in here during my tests. Let me know if you want to add any exception, I will add. And, just adding general exception is not a very good idea

longshuicy (approved):

Rashmil-1999 (approved):

PR #546, opened 2024-04-19
Fixed #519 removed unnecessary dependency

### LINKED ISSUE #519: Remove openssl library dependency from setup.py
There is no pip module called openssl so I think this can be safely removed from the setup.py. This was added so we could fix an issue with conda using the latest openssl.

navarroc:
openssl is not available through pip and breaks pip install of pyincore after 1.15.1. The openssl dependency is for conda installation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Fixed
+- Unnecessary dependency in setup.py [#519](https://github.com/IN-CORE/pyincore/issues/519)
+
 
 ## [1.18.0] - 2024-04-03
 
--- a/environment.yml
+++ b/environment.yml
@@ -20,4 +20,3 @@ dependencies:
   - rtree>=1.1.0
   - scipy>=1.11.3
   - shapely>=2.0.2
-  - openssl<=3.2.0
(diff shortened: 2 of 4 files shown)


ywkim312 (commented):
Whenever I install pyincore using pip in my local for test, I always remove this line. Why we need this for setup. Also without this, conda pyincore should be okay, right?

navarroc:
> Whenever I install pyincore using pip in my local for test, I always remove this line. Why we need this for setup? Also without this, pyincore should be okay, right?

Yes, we don't need this. Actually, if you try to pip install pyincore you cannot install any version after 1.15.1 because of this dependency got added. environment.yml already handles adding this for the conda build/install of pyincore so it's not needed in setup.py

ywkim312:
> > Whenever I install pyincore using pip in my local for test, I always remove this line. Why we need this for setup? Also without this, pyincore should be okay, right?
> 
> Yes, we don't need this. Actually, if you try to pip install pyincore you cannot install any version after 1.15.1 because of this dependency got added. environment.yml already handles adding this for the conda build/install of pyincore so it's not needed in setup.py

Thanks. approve

ywkim312 (approved):

longshuicy (commented):
Does other openssl matters? I know environment.yaml is for conda, but what about e.g. requirements.min?

image

navarroc:
> Does other openssl matters? I know environment.yaml is for conda, but what about e.g. requirements.min?
> 
> image

Let me try removing openssl altogether. We pinned the versions a while back because conda installing a higher version broken the github actions. Let me see what happens if I unpin the version in requirements.min.

navarroc:
Looks like whatever the bug was with openssl got fixed so I have removed the version pin in requirements.min and environment.yml. This should be ready for final review. If you want, I can run the requirements.sh script to update the versions.md file.

navarroc:
@longshuicy when you get a chance, can you take another look at this? Thanks!

longshuicy (approved):

PR #549, opened 2024-04-23
added documentation for ML CGE related modules

Rashmil-1999:
With this PR I added documentation for CoreCGEML, MlEnabledCgeSlc, and added a CGEMLFileUtil module which modifies the file by moving the functions in the class.


navarroc:
Does a reference to the new CGEMLFileUtil need to go in __init__.py in the pyincore folder? if it doesn't it might not be available to external packages like pyincore-incubator. I notice the other utils are there in that file

Rashmil-1999:
Ahh, that is what I thought, since it wasn't there in the root level I thought that we are good and it doesn't need the import. I didn't see we were importing it at the root level. Thanks :)

pyincore/utils/cge_ml_file_util.py
```diff
@@ -2,158 +2,160 @@
...
-) -> Tuple[Dict[str, List[str]], np.ndarray, np.ndarray, Dict[str, np.ndarray]]:
-    """parse_files Utility function to parse the input files
-
-    Parameters
-    ----------
-    model_filenames : Dict[str, str]
-        Dictionary containing the factor name and the path to the coefficient file
-    filenames : List[str]
-        Paths to the .csv files containing the base values. It has to be organized this order, starting from:
+class CGEMLFileUtil:
```
navarroc, line 10: You need to add another blank line before the class name, I see pep8 error about there only being one line between the class name and the logger = pyglobals.LOGGER line

navarroc (approved):

ywkim312 (approved):

PR #564, opened 2024-05-06
Fixed DFR3 service's out of index error

ywkim312:
Test this with road_dmg.ipynb after installing the PR locally by pip install . -U


pyincore/dfr3service.py
```diff
@@ -395,10 +395,17 @@ def _convert_properties_to_dict(rules, properties):
         matched_properties = {}
         # Handle legacy rules
         if isinstance(rules, list):
+            # If the rules are empty, return the matched properties
+            if not rules:
+                return matched_properties
```
longshuicy, line 400: I think if rules are empty, we should match building to that curve.
@navarroc do you recall if that was that the design?

navarroc: Yes, if there are no rules, then it  should be a match. Also, you can test this by running test_roaddamage.py, which still fails for me.

ywkim312: Can you give me some examples. I don't know how to make match building to the curve part.

navarroc: I think you make make line 399 be:

if not rules or rules == [None]:

Right now you are only checking if the list is empty. I think it's not empty, but contains "None" in this case.

ywkim312: Oh I got it. I modified it. I tested with the road damage and it passed the part of this line but didn't go until the end since the test file was removed from dev, like you mentioned in other PR.

navarroc: Can you re-upload the galveston roads and update the ID in the test?

navarroc: We can probably use the one from production for galveston and copy it to dev. Looks like there are two of them.

ywkim312:
This PR contains the change of the test_roaddamage.py but it should be the same one as the existing. Somehow, I can't remove this from the PR.

pyincore/dfr3service.py
ywkim312: @navarroc I only see this one 606dd68fbe94522d1cc034b2 but it is a csv file. Is this correct? Or can you specify dataset id in here?

navarroc: On Prod: 5f15d04f33b2700c11fc9c4e

navarroc:
Looks like the fragility needs distance, which this dataset has on production: 60ba6340aed73f24c5c5eaca

navarroc (approved):
test_roaddamage.py and the road_dmg.ipynb notebook both pass now

tests/pyincore/analyses/roaddamage/test_roaddamage.py
```diff
@@ -63,7 +63,7 @@ def run_with_base_class():
     # test galveston hurricane road failure
 
     # road inventory for Galveston island
-    road_dataset_id = "60ba5ee94219ee25a56b6999"
+    road_dataset_id = "664e5812efdc9f1ed5dc2f7f"
```
longshuicy, line 66: Could you also update the tests/pyincore/analyses/roaddamage/test_roaddamage_w_hazard_obj.py. It is using the same id which doesn't exist anymore.

longshuicy (changes requested):

tests/pyincore/analyses/roaddamage/test_roaddamage.py
ywkim312: updated

longshuicy (approved):

PR #565, opened 2024-05-08
560 renaming housing recovery to housing valuation recovery

longshuicy:
1. Use the sphinx deprecation decorator 
2. Found an elegant way for he old class to call the new one
3. Update the sphinx doc, also add the deprecation decorator

image


navarroc:
I think we'll need to add deprecated to the list of dependencies, right?

longshuicy:
> I think we'll need to add deprecated to the list of dependencies, right?

Hmm sorry I don't follow. Could you give me an example?

navarroc:
> > I think we'll need to add deprecated to the list of dependencies, right?
> 
> Hmm sorry I don't follow. Could you give me an example?

I had to install the "deprecated" library into my pip environment to run the test so we might need to add it to setup.py. Looking at the conda environment for 1.18.1, it is already there (probably part of the base conda package) or another dependency we have already pulls it in.

navarroc:
Basic tests pass (checked old output vs new output), the old analysis shows a deprecation message. I also checked/built the module documentation, looks good.

longshuicy:
> > > I think we'll need to add deprecated to the list of dependencies, right?
> > 
> > 
> > Hmm sorry I don't follow. Could you give me an example?
> 
> I had to install the "deprecated" library into my pip environment to run the test so we might need to add it to setup.py. Looking at the conda environment for 1.18.1, it is already there (probably part of the base conda package) or another dependency we have already pulls it in.

Ah i see! You are talking about the deprecated library. Yes I think I need to add it good catch!

longshuicy:
> I had to install the "deprecated" library into my pip environment to run the test so we might need to add it to setup.py. Looking at the conda environment for 1.18.1, it is already there (probably part of the base conda package) or another dependency we have already pulls it in.

I added the dependencies:
- Not sure how conda got the "deprecated" library. We are not explicitly listing it in the environment.yml... just to be safe I added there
- For pip, the package name seems to be capitalized "Deprecated"? https://pypi.org/project/Deprecated/

pyincore/analyses/housingvaluationrecovery/housingvaluationrecovery.py
```diff
@@ -0,0 +1,382 @@
+# This program and the accompanying materials are made available under the
```
navarroc, line 1: Copyright year is missing

pyincore/analyses/housingrecovery/housingrecovery.py
```diff
@@ -2,381 +2,18 @@
 # terms of the Mozilla Public License v2.0 which accompanies this distribution,
```
navarroc, line 2: Same here, missing copyright year

longshuicy: added

pyincore/analyses/housingvaluationrecovery/housingvaluationrecovery.py
longshuicy: added

pyincore/analyses/housingvaluationrecovery/housingvaluationrecoveryutil.py
```diff
@@ -0,0 +1,128 @@
+# This program and the accompanying materials are made available under the
```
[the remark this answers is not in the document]
longshuicy: added

navarroc (approved):

ywkim312 (approved):

PR #571, opened 2024-05-13
562 update building nonstructural damage to support flood

longshuicy:
This PR:
1. Refactor nonstructural building damage to accept one fragility key at time and run damage analysis
2. Add support for flood
3. Rewrite pytest example of AS and DS
4. Add pytest for flood; remove flood example from building damage 

Related PR:
incore-docs: https://github.com/IN-CORE/incore-docs/pull/386


pyincore/analyses/nonstructbuildingdamage/nonstructbuildingdamage.py
```diff
@@ -428,13 +348,13 @@ def get_spec(self):
                     'id': 'result',
                     'parent_type': 'buildings',
                     'description': 'CSV file of damage states for building non-structural damage',
-                    'type': 'ergo:nsBuildingInventoryDamageVer3'
+                    'type': 'ergo:buildingDamageVer6'
```
navarroc, line 351: Even though the structure is the same as building damage (limit state and damage states), do you think we should still have a non structural damage type? Just asking since otherwise we could get rid of this analysis altogether if it produces the same output as building damage and has the same output types. But I think there is some value in knowing that is is non structural damage because otherwise from the shape of the output there is no way to tell.

longshuicy: I agree. I will switch it back to the non structural type and bump the version up `nsBuildingInventoryDamageVer4`

navarroc:
If I set use_liquefaction and a liquefaction_dataset_id with the flood test, I get an error. I think we have to check to make sure the hazard type is EQ or ignore the liquefaction as we do in building damage.

navarroc (approved):
Looks good. Compared the combined damage before and after the change, results match. Also verified setting liquefaction with the flood hazard works correctly.

ywkim312 (commented):
I got this error when I ran test_buildingdamage_legacy
requests.exceptions.HTTPError: 404 Client Error: Not Found for url: https://incore-dev.ncsa.illinois.edu/data/api/datasets/6091d5a8daa06e14ee96d502
Could you check? test_buildingdamage ran okay

ywkim312:
All other tests were fine except test_buildingdamage_legacy

navarroc:
Non structural damage needs to support retrofit_strategy (like building damage). I left a comment on the docs PR, but we should either include that in this PR or open a bug to fix it in a separate PR.

pyincore/analyses/nonstructbuildingdamage/nonstructbuildingdamage.py
```diff
@@ -33,17 +34,24 @@ def run(self):
...
         building_set = self.get_input_dataset("buildings").get_inventory_reader()
 
+        # building retrofit strategy
+        retrofit_strategy_dataset = self.get_input_dataset("retrofit_strategy")
+
+        # mapping
+        dfr3_mapping_set = self.get_input_dataset("dfr3_mapping_set")
+
+        # Update the building inventory dataset if applicable
+        bldg_dataset, tmpdirname, _ = DatasetUtil.construct_updated_inventories(building_set,
```
navarroc, line 44: building_set needs to be a Dataset object, but on line 34 we call get_inventory_reader()

longshuicy: Yes I just pushed a fix.

longshuicy:
@navarroc @ywkim312 sorry for the delay, this is ready to be reviewed again. I also updated the incore-docs PR.

navarroc (approved):
Tested using test classes and https://github.com/IN-CORE/incore-docs/pull/394

PR #572, opened 2024-05-13
Rename Social Vulnerability to Social Vulnerability Score

ywkim312:
- [x] Need to add how to deprecate
- [x] Add deprecate to requirements
- [x] Add deprecate to setup
- [x] Need to update documentation
- [x] Need to figure out if we change the input data type
- e.g, social_vulnerability_demographic_factors or socialVulnerabilityFeatureAverages. If so, the dataset type should also be changed

### DIFF
--- a/pyincore/analyses/socialvulnerability/socialvulnerability.py
+++ b/pyincore/analyses/socialvulnerability/socialvulnerability.py
@@ -1,0 +1,19 @@
+# This program and the accompanying materials are made available under the
+# terms of the Mozilla Public License v2.0 which accompanies this distribution,
+# and is available at https://www.mozilla.org/en-US/MPL/2.0/
+
+from deprecated.sphinx import deprecated
+
+from pyincore.analyses.socialvulnerabilityscore import SocialVulnerabilityScore
+
+
+@deprecated(version='1.19.0', reason="This class will be deprecated soon. Use SocialVulnerabilityScore instead.")
+class SocialVulnerability():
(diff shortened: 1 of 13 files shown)


longshuicy (approved):
Besides the missing copyright year everything looks good.
Both old and new analysis runs and results are identical.
Approving but please add the missing copyright back. Thanks!

pyincore/analyses/socialvulnerability/socialvulnerability.py
```diff
@@ -1,163 +1,19 @@
-# Copyright (c) 2021 University of Illinois and others. All rights reserved.
```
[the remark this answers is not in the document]
ywkim312: Should be there now

pyincore/analyses/socialvulnerabilityscore/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) 2024 University of Illinois and others. All rights reserved.
```
navarroc, line 1: This should match the original copyright year, 2021

ywkim312: I have already updated. Maybe you're using the old one? Can you pull one more time?

ywkim312: oh my bad, changed it.

navarroc (approved):

PR #575, opened 2024-05-20
567 create building repair analysis

longshuicy:
Code wise it is working, but not sure scientifically if it makes sense.
This analysis takes the `recovery_rate` function in residential/commercial recovery analysis and calculates it without the delay factor. 
- Repair curves are matched to each building
- Given MCS output can determine the sample size N
- For a given building, at each sample (N sample in total), create N size random number and plugin the repair curve to calculate the repair time (N)
- This will result in a **N * N matrix** of repair time for each building


navarroc:
@jonglee1 who can we consult with on the logic for this analysis and whether it makes sense scientifically?

PR #576, opened 2024-05-20
557 rename monte carlo failure probability to monte carlo limit state probability

longshuicy:
There are a lot of "failure" keywords in the function name, i/o name, and internal variable names. Not sure should I change all of them to "limit state"? 
Some of them won't make sense e.g. "failure_state_keys" --> "limit_state_keys"?


pyincore/analyses/montecarlolimitstateprobability/__init__.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) 2019 University of Illinois and others. All rights reserved.
```
navarroc, line 1: minor nitpick - I think copyright should be 2024 maybe? I know the original was marked 2019, but maybe this is a dev discussion.

navarroc: Just leaving this here to close the loop. It was decided we should keep the original copyright.

docs/source/modules.rst
```diff
@@ -217,7 +217,12 @@ analyses/mlenabledcgeslc
 
 analyses/montecarlofailureprobability
 =====================================
-..  autoclass:: montecarlofailureprobability.montecarlofailureprobability.MonteCarloFailureProbability
+.. deprecated:: 1.19.0
+   This class will be deprecated soon. Use :class:`montecarlolimitstateprobability.MonteCarloLimitStateProbability` instead.
+
+analyses/montecarlolimitstateprobability
+=====================================
```
ywkim312, line 224: can you match the number of "=" in line 224 the same as the number of characters in line 223?

ywkim312: also, I just remembered the above things, so let's also check this again for the other ones after we create a release branch using the release branch

longshuicy: I matched everything in the module.rst and will do another check on release branch when it's there.

ywkim312 (approved):
code looks good. test run worked. approve

navarroc (approved):
Code runs without issue and compared to previous results. Also checked building the documentation.

PR #582, opened 2024-06-05
Rename nonstructural building damage

ylyangtw:


pyincore/analyses/nonstructbuildingdamage/nonstructbuildingdamage.py
```diff
@@ -4,379 +4,19 @@
...
 
-class NonStructBuildingDamage(BaseAnalysis):
-    """Computes non-structural structural building damage for an earthquake hazard.
-
-    Args:
-        incore_client (IncoreClient): Service authentication.
-
-    """
-
+@deprecated(version='1.19.0', reason="This class will be deprecated soon. Use TrafficFlowRecovery instead.")
```
longshuicy, line 13: typo "TrafficFlowRecovery" should be use BuildingNonStructDamage instead.

navarroc:
> Why did you rename the original folder? The original should stay the same path. Also, any reason not to include full name? It should be BuildingNonstructuralDamage for name and the path would be buildingnonstructuraldamage

Oops, I saw that wrong. But can we make the new one the full name instead of the abbreviated?

ylyangtw:
> > Why did you rename the original folder? The original should stay the same path. Also, any reason not to include full name? It should be BuildingNonstructuralDamage for name and the path would be buildingnonstructuraldamage
> 
> Oops, I saw that wrong. I thought you had renamed the original path but the original was abbreviated. But can we make the new one the full name instead of the abbreviated?

Yeah I wasn't sure what's the "full name" so I just referenced the old name. Do you mean `buildingnonstructuraldamage`?

longshuicy (approved):

ywkim312 (approved):

PR #583, opened 2024-06-05
Rename building damage to building structural damage 

ylyangtw:


longshuicy (approved):
Code looks good and run well. Compared some of the results and they are identical. Approved.

docs/source/modules.rst
```diff
@@ -14,6 +14,15 @@ analyses/buildingdamage
     :members:
 ..  autoclass:: buildingdamage.buildingutil.BuildingUtil
     :members:
+..  deprecated:: 1.19.0
+    This class will be deprecated soon. Use :class:`buildingstructuraldamage.BuildingStructuralDamage` instead.
+
+analyses/buildingstructuraldamage
+=======================
```
ywkim312, line 21: I am very sorry for this because I am not very sure, but sphinx makes rendering error that doesn't actually giving any error message but doesn't show the part when the number of "=" doesn't match with the title above. So could you match the length of "=" as the same  for the number of characters in line 20?

ywkim312 (approved):
Worked well except my comment. I approve but please address my comment.

docs/source/modules.rst
ylyangtw: Thanks for finding this out! I just pushed the change commit

PR #591, opened 2024-06-24
585 indp solver failed

longshuicy:
Turns out i just need to downgrade pyomo; scip is ok with higher version (9.0.1 on my machine).

To Test:

1. run the pytest test_indp.py
2. run the notebook in incore-docs https://github.com/IN-CORE/incore-docs/blob/develop/notebooks/indp.ipynb


navarroc:
Looks good and works on my machine with the downgraded pyomo (not sure why that didn't work before). One minor change though, can we make copies of the prod datasets on dev and update the test to use that instead of using production for testing?

tests/pyincore/analyses/indp/test_indp.py
```diff
@@ -28,7 +28,7 @@
 def run_with_base_class():
     dev_client = IncoreClient(pyglobals.INCORE_API_DEV_URL)
     prod_client = IncoreClient()
-    dev_datasvc = DataService(dev_client)
+    prod_datasvc = DataService(prod_client)
```
navarroc, line 31: Just following up here, when you get a chance, can you change this back to using data on dev? I think it's better if tests only use our dev instance, even if these aren't automatically ran.

longshuicy: Yes I was planning to do that but I lost track on this. Will do shortly!

longshuicy: It took me longer than I think to migrate all the datasets and mappings but they are done~~

navarroc (approved):

ylyangtw (approved):
Tested ok. Changes look good

PR #598, opened 2024-06-26
fix e731, e741, f401 and f403

longshuicy:
fix e731, e741, f401 and f403


ylyangtw (commented):
changes look good. 
I tried re-running the github action twice (6/26 and 6/28) and both failed for different reasons though

PR #599, opened 2024-06-26
I cannot removee e203, black keeps adding it back

longshuicy:
It turns out I cannot ignore E203 (space around colon) because black keeps adding it back. After some research, it seems that black is very opinionated and doesn't allow configuration for this specific case.

Another alternative is to write a custom shell script to strip those spaces and add it to the pre-commit hook. While this might feel a bit overkill at the moment, it is a viable solution based on the current limitations. Let me know if we would want to go that route. @navarroc


navarroc:
What is the issue about the space around the colon? Is it breaking something?

longshuicy:
> What is the issue about the space around the colon? Is it breaking something?

The pep8 rule E203 https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/ specify that no space is allowed around : However, black insist to add the space around : which conflicts the E203.

Without add it to ignore, black will automatically reformat to add the space back so it is never able to pass the pre-commit.

navarroc:
> > What is the issue about the space around the colon? Is it breaking something?
> 
> The pep8 rule E203 https://docs.astral.sh/ruff/rules/whitespace-before-punctuation/ specify that no space is allowed around : However, black insist to add the space around : which conflicts the E203.
> 
> Without add it to ignore, black will automatically reformat to add the space back so it is never able to pass the pre-commit.

How does Clowder get around this issue? I see it is setup to ignore e203 too.

ylyangtw (approved):
Looks good. I agree with ignoring E203.

navarroc (approved):

PR #605, opened 2024-07-16
Add limit/damage state

longshuicy:
Changes need to work with the port component damage in the incubator. Don't review/ merge this yet.


pyincore/models/fragilitycurveset.py
```diff
@@ -142,6 +147,7 @@ def calculate_damage_interval(
             ("earthquake+tsunami", "building", 3): FragilityCurveSet._3ls_to_4ds,
             ("tornado", "building", 3): FragilityCurveSet._3ls_to_4ds,
             ("hurricane", "building", 3): FragilityCurveSet._3ls_to_4ds,
+            ("hurricane", "warehouse", 3): FragilityCurveSet._3ls_to_4ds,
```
longshuicy, line 150: maybe this can be reused using the building damage; since warehouse is just building.

PR #606, opened 2024-07-30
Interdependent recovery of residential buildings and households

ylyangtw:
This PR is to support Lisa Wang's new analysis `Interdependent recovery of residential buildings and households`.

Two analyses have changes:
1. Residential building recovery
    - Add 2 additional output datasets, which are the same as commercial building recovery
2. Housing recovery sequential
    - Major idea is that zones can be generated based on either social vulnerability or household income.
    - Change `sv_generator` from hardcoded to 2 input datasets `zone_def_sv` and `zone_dev_hhinc`. To make sure analysis runs without inputs, default `zone_def_sv` and `zone_dev_hhinc` are provided.
    - Add one function to generate zones based on household income, and modify the model accordingly.

For testing, please run 
`test_residentialbuildingrecovery.py`
`test_housingrecoverysequential_sv.py`
`test_housingrecoverysequential_hhinc.py`


longshuicy (commented):
Code runs well besides the minor suggestions. Approved.

longshuicy (approved):

tests/pyincore/analyses/housingrecoveryserial/test_housingrecoverysequential_hhinc.py
```diff
@@ -0,0 +1,73 @@
...
+import pyincore.globals as pyglobals
+
+
+def run_with_base_class():
+    client = IncoreClient(pyglobals.INCORE_API_DEV_URL)
+
+    start = timer()
+
+    # dev Joplin testbed
+    population_dislocation = Dataset.from_file(
```
navarroc, line 18: Can we make this a dataset on dev? This is a pretty big file and is better served from the incore-dev and referenced by ID instead.

navarroc:
The test folder should match the analysis folder name. Since the folder is housingrecoverysequential, the test folder shouldn't be housingrecoveryserial it should be housingrecoverysequential. This looks like it was missed before. Can you rename that folder?

navarroc:
test_housingrecoverysequential_sv.py raises a value error about hhinc, is that expected?

pyincore/analyses/housingrecoverysequential/housingrecoverysequential.py
```diff
@@ -352,6 +416,7 @@ def housing_serial_recovery_model(
         result["guid"] = households_df["guid"]
         result["huid"] = households_df["huid"]
         result["Zone"] = households_df["Zone"]
+        result["Zone Indicator"] = households_df["Zone Indicator"]
```
navarroc, line 419: This can probably be shortened to match the 10 char limit for dbf/shapefile. Maybe zone_indic?

navarroc:
Do we need to open issues to add new semantic types for the outputs from residential building recovery?

navarroc:
> test_housingrecoverysequential_sv.py raises a value error about hhinc, is that expected?

@longshuicy Raising a value error if hhinc is not present prevents the analysis from running, should it do that? It looks like the population_dislocation dataset referenced in the test doesn't have that column, probably because the housing unit allocation dataset used to generate it didn't have that column. If hhinc should be there, then we should raise a value error and fix the dataset on dev. Since this isn't a test we execute as part of the pyincore testing, we could fix this later so we can get this merged.

ylyangtw:
> Do we need to open issues to add new semantic types for the outputs from residential building recovery?

Just opened an issue https://github.com/IN-CORE/pyincore/issues/611

ylyangtw:
@navarroc I have addressed all the comments. Thank you!

navarroc (approved):

PR #612, opened 2024-09-03
608 equity metrics calculation

longshuicy:
Equity Metric takes 3 input:
- HUA Household inventory with sociodemographic characterization (or any characterization for which
can divide into two groups)
- Scarce Resource (could be anything, in this case we use return time calculated from HHRS)
- A Division decision variable column name that can divide a given household into 2 groups (e.g. low income vs non low income, minority vs non-minority). In this case, we calculate the social vulnerability index (SVI)

I made the algorithm generic regardless of what scarce resource and division decision variable

---

To test:

run `tests/pyincore/analyses/equitymetric/test_equitymetric.py`

### DIFF
--- a/pyincore/analyses/equitymetric/equitymetric.py
+++ b/pyincore/analyses/equitymetric/equitymetric.py
@@ -10,7 +10,7 @@ from pyincore.analyses.equitymetric.equitymetricutil import EquityMetricUtil
 
 
 class EquityMetric(BaseAnalysis):
-    """Computes electric power infrastructure functionality.
+    """Computes equity metric.
     Args:
         incore_client: Service client with authentication info
     """
--- a/pyincore/analyses/equitymetric/equitymetric.py
+++ b/pyincore/analyses/equitymetric/equitymetric.py
@@ -6,6 +6,7 @@
 
 import numpy as np
 from pyincore import BaseAnalysis
+from pyincore.analyses.equitymetric.equitymetricutil import EquityMetricUtil
 
 
 class EquityMetric(BaseAnalysis):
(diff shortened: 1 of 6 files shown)


pyincore/analyses/equitymetric/equitymetric.py
```diff
@@ -0,0 +1,140 @@
...
+# terms of the Mozilla Public License v2.0 which accompanies this distribution,
+# and is available at https://www.mozilla.org/en-US/MPL/2.0/
+
+import numpy as np
+from pyincore import BaseAnalysis
+from pyincore.analyses.equitymetric.equitymetricutil import EquityMetricUtil
+
+
+class EquityMetric(BaseAnalysis):
+    """Computes electric power infrastructure functionality.
```
navarroc, line 13: Can you update the description for this analysis?

navarroc:
@longshuicy It's minor, but can you add the equitymetrics analysis to the modules.rst in the docs folder? Thanks!

navarroc (approved):

navarroc:
I noticed that there is a file written "scarce_resource" in the test folder when the analysis exits. Is that supposed to be there? I only see one output listed in the spec.

longshuicy:
> I noticed that there is a file written "scarce_resource" in the test folder when the analysis exits. Is that supposed to be there? I only see one output listed in the spec.

That's the example I'm preparing the input datasets. 

```
# prepare input dataset
    return_df = Dataset.from_data_service(
        "66d7763b43810e1298b0e8b1", datasvc
    ).get_dataframe_from_csv()
    scarce_resource_df = EquityMetricUtil.prepare_return_time_as_scarce_resource(
        return_df
    )
    scarce_resource = Dataset.from_dataframe(
        scarce_resource_df, "scarce_resource", data_type="incore:scarceResource"
    )
    ```

Since anything could be scarce resources, if anyone has brought in their own, they don't have the run above code to generate that input.

ywkim312:
The test run finished without error and did not create any result or any output. The code looks like it doesn't produce any output. This is okay, right?

longshuicy:
> The test run finished without error and did not create any result or any output. The code looks like it doesn't produce any output. This is okay, right?

It should produce 2 files. The .csv is the final output.
Could you double check how you run the test file? I saw you approved the notebook, assuming it should run file with jupyter notebook. 
Thanks!
image

ywkim312 (approved):

PR #615, opened 2024-09-17
change dict to list to respect order

longshuicy:
Anibal reported that when uploading the hurricane rasters with multiple demand types to the Incore services, the returned hazard intensity values differ from the local results. Upon investigation, I found that the order of the files, which should correspond to the demand types, was not being preserved due to the use of a dictionary. I have updated the implementation to use a list instead, which should resolve the issue.

---

To Test:

Run the pytest or monitor the GitHub Action to ensure it passes.


navarroc (approved):

ywkim312 (approved):
pytest in github action passed. Approve

PR #618, opened 2024-10-09
588 implement mlenabledcgejoplin

Rashmil-1999:
run the test_mlcgejoplin.py in tests


navarroc:
Looks good. Checked documentation build and ran test. One quick question, is there any description we can give it similar to the JoplinCGEModel class or the Salt lake CGE model class? I noticed both SLC and Joplin ML models have no description. It might be worth seeing if they can provide something.

navarroc (approved):

ywkim312 (approved):

PR #628, opened 2024-10-30
Fix cge core and joplin basevals

Rashmil-1999:


longshuicy (approved):
Code runs but I don't understand the science completely. Approving.

PR #639, opened 2024-11-27
Add default internal networks for each service

longshuicy:
Set up different internal networks for each services when "internal" flag is set to true.

### DIFF
--- a/pyincore/dataservice.py
+++ b/pyincore/dataservice.py
@@ -31,12 +31,27 @@ class DataService:
 
     def __init__(self, client: IncoreClient):
         self.client = client
-        self.base_url = urljoin(client.service_url, "data/api/datasets/")
-        self.files_url = urljoin(client.service_url, "data/api/files/")
-        self.base_earthquake_url = urljoin(
-            client.service_url, "hazard/api/earthquakes/"
-        )
-        self.base_tornado_url = urljoin(client.service_url, "hazard/api/tornadoes/")
+
+        if self.client.internal:
+            self.base_url = urljoin(
+                pyglobals.INCORE_INTERNAL_DATA_API_URL, "data/api/datasets/"
+            )
+            self.base_earthquake_url = urljoin(
+                pyglobals.INCORE_INTERNAL_HAZARD_API_URL, "hazard/api/earthquakes/"
+            )
+            self.base_tornado_url = urljoin(
+                pyglobals.INCORE_INTERNAL_HAZARD_API_URL, "hazard/api/tornadoes/"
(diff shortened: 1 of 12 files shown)


pyincore/dataservice.py
```diff
@@ -31,12 +31,27 @@ class DataService:
...
     def __init__(self, client: IncoreClient):
         self.client = client
-        self.base_url = urljoin(client.service_url, "data/api/datasets/")
-        self.files_url = urljoin(client.service_url, "data/api/files/")
-        self.base_earthquake_url = urljoin(
-            client.service_url, "hazard/api/earthquakes/"
-        )
-        self.base_tornado_url = urljoin(client.service_url, "hazard/api/tornadoes/")
+
+        if self.client.internal:
```
ywkim312, line 35: In the description, it says "when "internal" flag is set to true." How to set the internal flag?

longshuicy: You can pass in the flag when instantiate the client
`client = IncoreClient(internal=True)`

longshuicy: But I don't think you can test it since the network name is internal too

ywkim312: code looks okay to me, but if I can't test, should I just approve it?

navarroc:
I don't know if this is overkill, but we may at some point want to add a "local" option for the client for users that wants to test the service running on their local machine (e.g. localhost:8080) . Right now, if we set the client as internal and the url to localhost:8080, the URL set by the user will never used because by internal we mean internal to the cluster and we grab the internal URLs (incore-data-svc:8888, etc). I don't see a way to ever set localhost:8080 and have it use it anymore, unless I missed something. 

That being said, this looks good. Do we want to create an RC and test it on the test cluster to verify all the URLs are working?

longshuicy:
> I don't know if this is overkill, but we may at some point want to add a "local" option for the client for users that wants to test the service running on their local machine (e.g. localhost:8080) . Right now, if we set the client as internal and the url to localhost:8080, the URL set by the user will never used because by internal we mean internal to the cluster and we grab the internal URLs (incore-data-svc:8888, etc). I don't see a way to ever set localhost:8080 and have it use it anymore, unless I missed something.
> 
> That being said, this looks good. Do we want to create an RC and test it on the test cluster to verify all the URLs are working?

Good idea. I was slightly worried that how to use the localhost:8080. It's not an overkill, I can add that quickly.

ywkim312 (approved):

navarroc (approved):

PR #642, opened 2025-01-07
641 add ml enabled galveston cge

Rashmil-1999:
To test, you must install this branch and run `python tests/pyincore/analyses/mlenabledcgegalveston/test_mlcgegalveston.py`


navarroc:
This is just for Galveston Island, right? If so, should this be called mlenabledcgegalvestonisland?

ywkim312:
When I ran the pytest, it said it was empty suite, and I believed that it was because my pytest did not detect any tests in your file. So maybe you want to update the pytest file to something like

```
from pyincore import IncoreClient, globals as pyglobals
from pyincore.analyses.mlenabledcgegalvestonisland import MlEnabledCgeGalvestonIsland

def test_mlcgegalvestonisland_run():
    client = IncoreClient(pyglobals.INCORE_API_DEV_URL)
    mlcgegalveston = MlEnabledCgeGalvestonIsland(client)
    
    sector_shocks = "677c0c346a42fa7faea05dac"
    
    mlcgegalveston.load_remote_input_dataset("sector_shocks", sector_shocks)
    mlcgegalveston.set_parameter("result_name", "test_galvestonisland_mlcge_result")
    
    mlcgegalveston.run_analysis()
```

Then you might also want to add the assertion something like
```
assert result is True  # or some other meaningful check
```

longshuicy (approved):
LGTM

navarroc (approved):
Tested and compared with previous model results.

Issue #646, opened 2025-02-21
Update Description in the Spec for the following analyses


Rashmil-1999:
1. CommercialBuildingRecovery
2. CumulativeBuildingDamage
3. EpfDamage
4. EpfRepairCost
5. EpfRestoration
6. EpnFunctionality
7. GalvestonCGEModel
8. GasFacilityDamage
9. HousingRecoverySequential
10. HousingUnitAllocation

PR #653, opened 2025-03-05
Update analysis spec descriptions

navarroc:
Reviewed and updated the spec description for inputs, parameters and each analysis listed in #646


Rashmil-1999 (approved):
Looks good, only one minor thing, can we write "CSV" instead of "csv"

longshuicy (approved):

PR #655, opened 2025-03-19
Fixes #654 - changed hazard service earthquake post method to allow soil type dataset

### LINKED ISSUE #654: Update hazard service earthquake values method to allow passing soil type dataset
The hazard service post_earthquake_hazard_values doesn't allow a user to specify the dataset containing the soil type information. The Java service endpoint already supports setting the dataset ID so we just need to allow the python method to pass this to the service.

navarroc:
I noticed that the Java service endpoint for earthquake allows a user to specify the soil type information, but the python method doesn't have any way to add the dataset id so the service will amplify the hazard based on the soil information at the location specified (otherwise it defaults to whatever soil type is specified when creating the earthquake).

Rather than adding another parameter to the method and then updating the earthquake python model, I added this information to the payload and checked for it in the service to see if it's present. I am not sure if this is the best way to do this so I am open to other options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [Unreleased]
+
+### Changed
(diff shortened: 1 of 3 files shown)


longshuicy (approved):

navarroc:
You should be able to test this with the building structural damage test by setting the site_class_dataset_id for the first test (Shelby county). Here is a site class file on dev - 5a284f56c7d30d13bc0824f6

ywkim312 (approved):
Test worked well.

PR #663, opened 2025-04-29
WIP: 657 implement ml enabled cge seaside

longshuicy:


navarroc:
It looks like the file DS_base_val.csv expects to have a first column of index values. If you compare the seaside file to https://github.com/IN-CORE/pyincore/blob/develop/pyincore/analyses/mlenabledcgejoplin/DS_base_val.csv you'll see what I mean.

After I made the change, the next error is it's complaining that 
```
ValueError: Sector AGRI not found in the sector shocks file with
 0      SERV
1    HEALTH
2      UTIL
3     CONST
4     MANUF
5        HS
6        AG
7     ACCOM
8      REST
9    RETAIL
```

navarroc (approved):

PR #668, opened 2025-06-17
Fixes #667 - network analyses check for network dataset input

### LINKED ISSUE #667: Analyses that require network dataset should check if a user has set a network dataset
Analyses should check if a network dataset is set by the user and if so, use the object as-is instead of trying to re-instantiate it again. If the type is Dataset, then it should try to instantiate a network dataset.

navarroc:
This fix allows a user to specify a local network dataset as input to the analysis. The analysis checks to see if the input is an instance of a network dataset and if it is, uses it as is. If it's a Dataset type, then it will attempt to load the network dataset from the Dataset. Previously, it always assumed it was a Dataset and tried to load it into a NetworkDataset.


longshuicy (approved):
Works well approved.

ywkim312 (approved):

PR #672, opened 2025-07-29
change capital shocks analysis to output missing guid sectors too

Rashmil-1999:
This PR will tweak the capital shocks analysis to also output sectors that do not have guids in the building to sector mapping file with unit shocks.


longshuicy (approved):

Rashmil-1999:
To test this, install this branch and run the new mlcge lumberton notebook here https://github.com/IN-CORE/incore-docs/pull/451 *NOTE* the ML CGE wont work since the code is not present on the installed branch so only run till the capital shocks

